PR #5, opened 2015-12-04
[bug-fix] [admin-menu] Wrapped upgrade menu in a conditional to check for paid plans - fixes #4

### LINKED ISSUE #4: Upgrade menu added without paid plans
Summary:
If I don't have paid plans for my plugin, then I shouldn't see the Upgrade menu added.

You can see I've set "has_paid_plans" to false; however, it still adds the submenu.

code

upgrade_submenu

I'll be submitting a commit for this issue soon. Please confirm that this is a bug and I'm not crazy. :)

irkanu:
I wanted to leave a few notes here:

My conditional checks the _has_paid_plans attribute; however, the conditionals only a few lines above this check using a _has_addons function.

I think these should be consistent throughout the code - for example, maybe we could benefit from having a _has_paid_plans function? I'm not sure how you want to keep that structured, but it was kind of confusing having some internal functions using _has_xyz returning their _has_xyz attribute.

Let me know if I can be of any help. :+1:

### DIFF
--- a/freemius/includes/class-freemius.php
+++ b/freemius/includes/class-freemius.php
@@ -4646,18 +4646,20 @@
 					}
 
 					// Add upgrade/pricing page.
-					$this->add_submenu_item(
-						( $this->is_paying() ? __fs( 'pricing' ) : __fs( 'upgrade' ) . '&nbsp;&nbsp;&#x27a4;' ),
-						array( &$this, '_pricing_page_render' ),
-						$this->get_plugin_name() . ' &ndash; ' . __fs( 'pricing' ),
-						'manage_options',
-						'pricing',
-						array( &$this, '_clean_admin_content_section' ),
-						WP_FS__LOWEST_PRIORITY,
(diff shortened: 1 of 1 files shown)


irkanu:
Oops, I just noticed this was pulled to master. Let me know if this should be pulled to dev.

fs-contributor:
Hey @irkanu, that's our first pull-request - #exciting! Thanks for spending the time to check the code. I'm not sure if you noticed, but the last param of `add_submenu_item()` is `$show_submenu`, and `$this->has_paid_plan()` is part of the last param provided to the call. So basically, if the plugin doesn't have any paid plans the submenu item won't be visible. Am I missing something?

`_has_paid_plans` is a private member of the `Freemius` class. There is a public function named `has_paid_plan()` which you can use from outside. The best way to see all the public methods that would always be back compatible is by checking the abstract class `Freemius_Abstract` at `class-freemius-abstract.php`.

Let me know if you still think that a pull request is required.

Issue #7, opened 2015-12-07
Loading image on VVV does not work


brianhogg:
From https://github.com/Freemius/wordpress-sdk/issues/3 the instructions to load the plugin image does not work if the local development environment is not running on actual localhost.  In my VVV environment `$_SERVER['REMOTE_ADDR']` is `192.168.50.1` so this line in freemius/config.php will never be true:

```
	define( 'WP_FS__IS_LOCALHOST', ( substr( $_SERVER['REMOTE_ADDR'], 0, 4 ) == '127.' || $_SERVER['REMOTE_ADDR'] == '::1' ) );
```

and `WP_FS__IS_LOCALHOST` needs to be true for the icon to be downloaded locally in template/plugin-icon.php

brianhogg:
Also if the icon is not downloaded the HTML generated has the plugin's img src as a broken image rather than the plugin default image

fs-contributor:
Added `WP_FS__LOCALHOST_IP` constant to `config.php` for local environments with custom local IP + if from some reason the fails to fetch the plugin's icon, fall backs to generic local icon.

Ref: [f8df74e] [22d35aa]

PR #17, opened 2016-01-19
Little fix on config.php

Mte90:
In this way I can define WP_FS__SKIP_EMAIL_ACTIVATION in wp-config.php.
Also the vvv ip is 192.168.50.4 https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/Vagrantfile#L122

### DIFF
--- a/config.php
+++ b/config.php
@@ -39,7 +39,7 @@
 	 */
 	if ( ! defined( 'WP_FS__LOCALHOST_IP' ) ) {
 		// VVV default public network IP.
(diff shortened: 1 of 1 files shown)


vovafeldman:
`192.168.50.4` was the original value, but then @brianhogg said that it's `192.168.50.1` on his VVV environment [Issue #7] :-)

Mte90:
The actual official ip is `192.168.50.4` and i think that that need to set in the sdk, is a constant so for the people that have different is not a problem change.

PR #20, opened 2016-01-26
Fix Multiple Submenu Items in Common Parent Menu

majick777:
Setting to a static value here (10000) causes conflicts when multiple instances of this function are called on a common parent menu... 
ie. for branding purposes all my plugins are under the same admin parent menu, and the 10000 value kept overriding the earlier submenus set to the same value, causing them to disappear entirely from the $submenu array and thus the page.
Setting this to the next available (highest submenu item array key) value plus one instead fixes the problem (thoroughly tested here.)
Note: this is really a one-line commit, there are only two commits here because the first edit got mangled.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -1,4 +1,4 @@
-'".WP_PLUGIN_URL."/".$vtaggerdir."/tick.png'<?php
+<?php
 	/**
 	 * @package     Freemius
 	 * @copyright   Copyright (c) 2015, Freemius, Inc.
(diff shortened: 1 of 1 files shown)


vovafeldman:
Thanks for the PR @majick777! Your code assumes that `$top_level_menu` keys are numerically ordered and I think that it's not the case. If I'm right, a simple `max()` will resolve it. If I'm missing something please let me know.

majick777:
`$top_level_menu` is numerically ordered from my debugging, as it is already the submenu list for that top level menu accessed from `$submenu['toplevelmenuslug']` and the numeric keys are thus the actual submenu display order. I think using `max()` gives the highest array value rather than highest key (order.) An alternative might be to use `$i = max(array_keys($top_level_menu)) + 1;`

vovafeldman:
My suggestion:
`$i = max( 10000, max( array_keys( $top_level_menu ) ) + 1 );`

Try that in your environment and if it works for you update the Pull Request, and I'll merge it to the master.

PR #24, opened 2016-01-30
Added new actions to the permissions list on the opt in.

danieliser:
Also includes the fix needed to show support forum link whether the user registered or not.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -4977,21 +4977,33 @@
 			ksort( $top_level_menu );
 		}
 
+		/**
+		 * Displays the Support Forum link when enabled.
+		 *
+		 * Can be filtered like so:
+		 *
+		 *  function _fs_show_support_menu( $is_visible, $menu_id ) {
+		 *      if ( 'support' === $menu_id ) {
+		 * 		    return _fs->is_registered();
+		 * 		}
+		 * 		return $is_visible;
+		 * 	}
+		 * 	_fs()->add_filter('is_submenu_visible', '_fs_show_support_menu', 10, 2);
+		 *
+		 */
 		function _add_default_submenu_items() {
 			if ( ! $this->is_on() ) {
 				return;
 			}
 
-			if ( $this->is_registered() ) {
-				if ( $this->_menu->is_submenu_item_visible( 'support' ) ) {
-					$this->add_submenu_link_item(
-						$this->apply_filters( 'support_forum_submenu', __fs( 'support-forum', $this->_slug ) ),
-						$this->apply_filters( 'support_forum_url', 'https://wordpress.org/support/plugin/' . $this->_slug ),
-						'wp-support-forum',
-						'read',
-						50
-					);
-				}
+			if ( $this->_menu->is_submenu_item_visible( 'support' ) ) {
+				$this->add_submenu_link_item(
+					$this->apply_filters( 'support_forum_submenu', __fs( 'support-forum', $this->_slug ) ),
+					$this->apply_filters( 'support_forum_url', 'https://wordpress.org/support/plugin/' . $this->_slug ),
+					'wp-support-forum',
+					'read',
+					50
+				);
 			}
 		}
 
--- a/includes/fs-core-functions.php
+++ b/includes/fs-core-functions.php
@@ -308,4 +308,36 @@
 		fclose( $fp );
 	}
 
+    /* General Utilities
+    --------------------------------------------------------------------------------------------*/
+
+    /**
+     * Sorts an array by the value of the priority key.
+     *
+     * @author Daniel Iser (@danieliser)
+     * @since  1.1.7
+     *
+     * @param $a
+     * @param $b
+     *
+     * @return int
+     */
+    function fs_sort_by_priority( $a, $b ) {
+
+        // If b has a priority and a does not, b wins.
+        if ( ! isset( $a['priority'] ) && isset( $b['priority'] ) ) {
+            return -1;
+        }
+        // If b has a priority and a does not, b wins.
+        elseif ( isset( $a['priority'] ) && ! isset( $b['priority'] ) ) {
+            return 1;
+        }
+        // If neither has a priority or both priorities are equal its a tie.
+        elseif ( ( ! isset( $a['priority'] ) && ! isset( $b['priority'] ) ) || $a['priority'] === $b['priority'] ) {
+            return 0;
+        }
+
+        // If both have priority return the winner.
+        return ( $a['priority'] < $b['priority'] ) ? - 1 : 1;
+    }
 
--- a/includes/fs-core-functions.php
+++ b/includes/fs-core-functions.php
@@ -326,11 +326,11 @@
 
         // If b has a priority and a does not, b wins.
         if ( ! isset( $a['priority'] ) && isset( $b['priority'] ) ) {
-            return -1;
+            return 1;
         }
         // If b has a priority and a does not, b wins.
         elseif ( isset( $a['priority'] ) && ! isset( $b['priority'] ) ) {
-            return 1;
+            return -1;
         }
         // If neither has a priority or both priorities are equal its a tie.
         elseif ( ( ! isset( $a['priority'] ) && ! isset( $b['priority'] ) ) || $a['priority'] === $b['priority'] ) {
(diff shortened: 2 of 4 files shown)


danieliser:
Here is usage example of the action. https://github.com/PopupMaker/Popup-Maker/blob/feature/175/includes/class-pum-freemius.php#L85

vovafeldman:
1. I will merge both templates, `connect.php` and `pending-activation.php`, since they are almost identical.
2. Looking for the long run, the actions you added are great but don't allow injecting permissions in between. And since we making sure the SDK is always backward competible, I don't want to add something that I know for sure will be changed. So if you can spend the time turning it to a filter that receives an array of permissions and then renders them, that would be great. If not, we'll add it to the TODOs and push it future releases.

danieliser:
@vovafeldman - You are a mind reader. I just didn't know how you would feel about me making more than basic changes. Here is what I had envisioned.

            array(
                    'profile'    => array(
                            'icon'     => 'admin-users',
                            'label'    => __fs( 'permissions-profile', $slug ),
                            'desc'     => __fs( 'permissions-profile_desc', $slug ),
                            'priority' => 5,
                    ),
                    'site'       => array(
                            'icon'     => 'wordpress',
                            'label'    => __fs( 'permissions-site', $slug ),
                            'desc'     => __fs( 'permissions-site_desc', $slug ),
                            'priority' => 10,
                    ),
                    'newsletter' => array(
                            'icon'     => 'email-alt',
                            'label'    => __fs( 'permissions-newsletter', $slug ),
                            'desc'     => __fs( 'permissions-newsletter_desc', $slug ),
                            'priority' => 15,
                    ),
            );

On render it will be sorted by priority.

vovafeldman:
It would be easier to put the priority as the index key instead of a property, and also instead of _icon_ using _icon_class_, otherwise the developer is limited to _dashicons_.
```php
  array(
                5    => array(array(
                        'id'     => 'profile',
                        'icon-class'     => 'dashicons dashicons-admin-users',
                        'label'    => __fs( 'permissions-profile', $slug ),
                        'desc'     => __fs( 'permissions-profile_desc', $slug ),
                )),
                ...
        );
```

If you are good with that and can implement it that would be great :+1:

danieliser:
@vovafeldman - The reason I would do the id as a key is priorities change, ids do not, so if you wanted to unset a specific one then you can just do unset ( $array['profile'] ); rather than having to lookup the keys. Also the sorting is rather simple with the priority method, a lot of built in arrays are sorted the same way.

Also 2 rows can have the same priority, but not the same array key.

Your call, just my 2 cents.

vovafeldman:
What would be an efficient way to sort it by priority? I'm probably missing something.

danieliser:
Like this.

	uasort( $array, 'sort_by_priority' );
	function sort_by_priority( $a, $b ) {
		if ( ! isset( $a['priority'] ) || ! isset( $b['priority'] ) || $a['priority'] === $b['priority'] ) {
			return 0;
		}
		return ( $a['priority'] < $b['priority'] ) ? - 1 : 1;
	}

danieliser:
That can be simplified even further. in the sort function, but I don't enfoce the priority argument so I have to check for it there.

danieliser:
@vovafeldman What do you think of this, before I commit it. Also where do you want the permission list array generated? (in the freemius class? method name preference?)

    // Set core permission list items.
    $permissions = array(
        'profile'    => array(
            'icon'     => 'dashicons dashicons-admin-users',
            'label'    => __fs( 'permissions-profile' ),
            'desc'     => __fs( 'permissions-profile_desc' ),
            'priority' => 5,
        ),
        'site'       => array(
            'icon'     => 'dashicons dashicons-wordpress',
            'label'    => __fs( 'permissions-site' ),
            'desc'     => __fs( 'permissions-site_desc' ),
            'priority' => 10,
        ),
        'events'       => array(
            'icon'     => 'dashicons dashicons-admin-plugins',
            'label'    => __fs( 'permissions-events' ),
            'desc'     => __fs( 'permissions-events_desc' ),
            'priority' => 20,
        ),
    );

    // Add newsletter permissions if enabled.
    if ( $fs->is_permission_requested( 'newsletter' ) ) {
        $permissions['newsletter'] = array(
                'icon'     => 'dashicons dashicons-email-alt',
                'label'    => __fs( 'permissions-newsletter' ),
                'desc'     => __fs( 'permissions-newsletter_desc' ),
                'priority' => 15,
        );
    }

    // Allow filtering of the permissions list.
    $permissions = $fs->apply_filters( 'permission_list', $permissions );

    // Sort by priority.
    uasort( $array, 'sort_by_priority' );
    
    if ( ! empty( $permissions ) ) : ?>
	<div class="fs-permissions">
		<a class="fs-trigger" href="#"><?php _efs( 'what-permissions', $slug ) ?></a>
		<ul><?php
            foreach( $permissions as $id => $permission ) : ?>
			<li id="fs-permission-<?php esc_attr_e( $id ); ?>" class="fs-permission fs-<?php esc_attr_e( $id ); ?>">
				<i class="<?php esc_attr_e( $permission['icon'] ); ?>"></i>
				<div>
					<span><?php esc_html_e( $permission['label'] ); ?></span>
					<p><?php esc_html_e( $permission['desc'] ); ?></p>
				</div>
			</li>
            <?php endforeach; ?>
		</ul>
	</div>
    <?php endif; ?>

danieliser:
PS Now you can see how priority works efficiently since the newsletter is added after but listed in the middle.

vovafeldman:
It looks good. The only things that I would change are:

1. Rename the property from `icon` to `icon-class`.
2. Instead of adding custom sorting function that will only be used for the permissions (at least now), I'd use a different structure of the permissions and then use `ksort` instead:

```php
// Set core permission list items.
$permissions = array(
    5    => array(array(
        'id'               => 'profile',
        'icon-class'     => 'dashicons dashicons-admin-users',
        'label'    => __fs( 'permissions-profile' ),
        'desc'     => __fs( 'permissions-profile_desc' ),
    )),
    ...
);

// Add newsletter permissions if enabled.
if ( $fs->is_permission_requested( 'newsletter' ) ) {
    $permissions[15][] = array(
            'id'          => 'newsletter',
            'icon-class'     => 'dashicons dashicons-email-alt',
            'label'    => __fs( 'permissions-newsletter' ),
            'desc'     => __fs( 'permissions-newsletter_desc' ),
    );
}

// Allow filtering of the permissions list.
$permissions = $fs->apply_filters( 'permission_list', $permissions );

// Sort by priority.
ksort( $permissions );

if ( ! empty( $permissions ) ) : ?>
<div class="fs-permissions">
    <a class="fs-trigger" href="#"><?php _efs( 'what-permissions', $slug ) ?></a>
    <ul><?php
        foreach( $permissions as $priority => $permission ) : ?>
        <?php $id = $permission['id'] ?>
        <li id="fs-permission-<?php esc_attr_e( $id ); ?>" class="fs-permission fs-<?php esc_attr_e( $id ); ?>">
            <i class="<?php esc_attr_e( $permission['icon-class'] ); ?>"></i>
            <div>
                <span><?php esc_html_e( $permission['label'] ); ?></span>
                <p><?php esc_html_e( $permission['desc'] ); ?></p>
            </div>
        </li>
        <?php endforeach; ?>
    </ul>
</div>
<?php endif; ?>
```

I don't see any particular benefit to use the 'id' as the key vs. 'priority'. What do you think?

danieliser:
@vovafeldman - My immediate thought is nested nested arrays isn't exactly a clean solution.

The priority sort function could be re-used anywhere in your sdk you wanted. Menu items could be sorted for instance.

How would the unset functionality look if I wanted to remove one? No real way to remove a specific one without looping over them all to check for id matches.

Things to think about is all. I mean its your codebase so I will rework it that way, but I usually opt for simplicity and readability.

vovafeldman:
The `unset` is a valid point. I'm not throwing arguments because it's **my code** :) I actually love the involvement/contribution of others. I just want to make sure we take good design decisions and don't end up with spaghetti code.

I accept your suggestion: 
1. Add the sorting function into `fs-core-functions.php` and name it `fs_sort_by_priority` (add your info in the PHPDoc and use `@since 1.1.7`.
2. Use `icon-class` instead of `icon`.
3. Test it locally.
4. I'll merge :+1:

danieliser:
Awesome. Will have it pushed up as soon as I test it. That should close out my freemius changeset for Popup Maker as well.

danieliser:
Do you want the priorities array created inside the freemius class somewhere as a method rather than inline? I assumed that would be the case. Thoughts on name?

vovafeldman:
I would keep that in the template for now.

danieliser:
@vovafeldman Check it. Tested working perfectly.

PR #28, opened 2016-02-07
Update fs-plugin-functions.php

wpalchemist:
If the only plan is a free plan, display a "download" button instead of a "purchase" button.

### DIFF
--- a/includes/fs-plugin-functions.php
+++ b/includes/fs-plugin-functions.php
@@ -220,12 +220,16 @@
 								<?php $first = false; endforeach ?>
 						<?php endif ?>
 					</ul>
-					<?php echo ' <a class="button button-primary right" href="' . esc_url( add_query_arg( array(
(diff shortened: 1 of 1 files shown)


vovafeldman:
Did you test it? From the first look, I don't think the download link will work since the add-on is not in the repo. Or do you plan to host it there?

wpalchemist:
Yes, I tested and it is working fine on my site.  The plugin is hosted in the repo, so the download link does work.

wpalchemist:
Perhaps there should be an option somewhere in the Freemius dashboard for whether or not a plugin is in the repo.

vovafeldman:
Got it. We'll extend that to work in both situations. Thanks!

vovafeldman:
Integrated in e83ee0aa56249b95c652923fa77ed36f72297113

PR #60, opened 2016-06-06
PHP Tidy and PHPCBF

ajaydsouza:
Ran PHP Tidy and PHPCBF with WordPress Coding Standards on all the files. This fixes the automatically fixable WordPress standard errors generated by running PHPCS


GaryJones:
0.12.0 of WPCS is due out tomorrow, and a lot has changed since 0.9.0, so it may be worth re-starting this PR.

PR #61, opened 2016-06-09
"insufficient permissions" notice on upgrade; missing top level menu

intoxstudio:
1.
If the user has skipped opt-in and later decides to upgrade, he will be met by a permission notice when being redirected to the Account page.

This is because the Account is set after actions have run.

2.
Before the user decides to opt-in, the top level menu item is not displayed.
When new overriding menu item is added on the same position, it is immediately removed again by unsetting that position.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -6044,8 +6044,8 @@
 			if ( ! $this->has_api_connectivity() && ! $this->enable_anonymous() ) {
 				$this->_menu->remove_menu_item();
 			} else {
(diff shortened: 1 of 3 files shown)


vovafeldman:
- I managed to reproduce bug n.1, and your PR does resolve it.
- I couldn't reproduce bug n.2, maybe it's related to the WP version. Though your tweak in the order makes total sense.
- Your changes will be merged into the `develop` branch, and later to `master`.

Thanks for contributing @intoxstudio!

PR #63, opened 2016-06-10
Guard against freemius menu index error

cyberhobo:
A suggested fix for issues reported to Postmatic - looks like it should be safe for all.


vovafeldman:
Thanks @cyberhobo! It's already in the upcoming release 👍

cyberhobo:
Ah, I should have checked the develop branch. Thanks!

Issue #67, opened 2016-07-17
Leverage the WordPress.org plugins API to fetch the URL of the plugin's icon.


vovafeldman:
Currently `/templates/plugin-icon.php` tries to fetch 128x128 & 256x256 icons from the SVN using HTTP, and stores the 1st version that returns a real image. Apparently, it's possible to fetch the icon from the plugins API by adding `icons` as one of the `fields` params in the querystring:

`http://api.wordpress.org/plugins/info/1.0/{slug}.json?fields=banners,icons`

bfintal:
You can also use this for a smaller result size:

	$payload = array(
		'action' => 'plugin_information',
		'request' => serialize( (object) array(
			'slug' => 'page-builder-sandwich',
			'fields' => array(
				'tags' => false,
				'icons' => true,
				'sections' => false,
				'description' => false,
				'tested' => false,
				'requires' => false,
				'rating' => false,
				'downloaded' => false,
				'downloadlink' => false,
				'last_updated' => false,
				'homepage' => false,
				'tags' => false,
				'compatibility' => false,
				'ratings' => false,
				'added' => false,
				'donate_link' => false,
			),
		) ),
	);
	$body = wp_remote_post( 'http://api.wordpress.org/plugins/info/1.0/', array( 'body' => $payload ) );

	var_dump( unserialize( $body['body'] )->icons );

PR #68, opened 2016-07-17
Feature/fs theme mode

vovafeldman:


includes/class-freemius.php
```diff
@@ -240,7 +240,7 @@ private function __construct( $slug ) {
 
 			$this->_plugin_main_file_path = $this->_find_caller_plugin_file();
 			$this->_plugin_dir_path       = plugin_dir_path( $this->_plugin_main_file_path );
-			$this->_plugin_basename       = plugin_basename( $this->_plugin_main_file_path );
+			$this->_plugin_basename       = $this->_get_plugin_basename( $this->_plugin_main_file_path );
```
vovafeldman: @fajardoleo When a function is actually private, there's no need to add the '_' (underscore) prefix. For example in:
`private function _get_plugin_basename`

The underscore should be added to functions that supposed to be private (not used outside the SDK), but we had to make them public due to different reasons. For example, when the function is called with a filter/action, there's no way to keep it private.

You might find some functions that have the `private` directive but also have an underscore, it's my mistake.

includes/class-freemius.php
```diff
@@ -275,14 +275,65 @@ private function __construct( $slug ) {
 				FS_Api::clear_cache();
 			}
 
-			$this->_register_hooks();
+			if ( $this->_caller_is_plugin() ) {
```
vovafeldman: @fajardoleo `_register_hooks()` hooks some methods that are relevant both for plugins and themes. Therefore, you cannot just skip it when it's a theme.

Therefore, execute that method in both cases, and wrap the hooks that related only to plugins with a relevant `if` statement.

includes/class-freemius.php
```diff
@@ -275,14 +275,65 @@ private function __construct( $slug ) {
...
 
 		/**
+		 * Checks if the caller is a plugin. If it is not a plugin, then it should be a theme.
+		 *
+		 * @author Leo Fajardo (leorw)
+		 * @since  1.2.0
+		 *
+		 * @return bool
+		 */
+		private function _caller_is_plugin() {
```
vovafeldman: @fajardoleo please refactor to `is_caller_plugin`.

includes/class-freemius.php
```diff
@@ -275,14 +275,65 @@ private function __construct( $slug ) {
...
+
+		/**
+		 * @author Leo Fajardo (leorw)
+		 * @since  1.2.0
+		 *
+		 * @param  string $file_path
+		 *
+		 * @return string
+		 */
+		private function _get_plugin_basename( $file_path ) {
```
vovafeldman: @fajardoleo refactor to `get_plugin_basename`

includes/class-freemius.php
```diff
@@ -516,7 +567,11 @@ private function _find_caller_plugin_file() {
 				}
 			}
 
-			$plugin_file = fs_find_caller_plugin_file();
+			if ( $this->_caller_is_plugin() ) {
+				$plugin_file = fs_find_caller_plugin_file();
+			} else {
+				$plugin_file = dirname( WP_FS__DIR ) . '/functions.php';
```
vovafeldman: @fajardoleo I think this code is incorrect for cases when the newest SDK is running from a plugin, but currently, we are running it from a theme. The reason it will happen is because `WP_FS__DIR` is a reference to the SDK location, not to the context plugin/theme.

includes/entities/class-fs-plugin.php
```diff
@@ -84,6 +88,17 @@ function is_addon() {
...
 
+		/**
+		 * Returns "true" if the type is not "theme".
+		 *
+		 * @author Leo Fajardo (leorw)
+		 *
+		 * @return bool
+		 */
+		function is_plugin() {
+			return ( 'theme' !== $this->type );
```
vovafeldman: @fajardoleo for backward compatibility I would use:
`return ( ! isset($this->type) || 'theme' !== $this->type );`

includes/class-freemius.php
```diff
@@ -594,11 +649,13 @@ function _add_deactivation_feedback_dialog_box() {
 		 * @return array The uninstall reasons for the specified user type.
 		 */
 		function _get_uninstall_reasons( $user_type = 'long-term' ) {
+			$plugin_type = ( $this->_plugin->is_plugin() ? 'plugin' : 'theme' );
```
vovafeldman: @fajardoleo I would add a helper function to the plugin entity class and use it instead:

`function get_type() {
    return $this->is_plugin() ? 'plugin' : 'theme';
}
`

includes/class-freemius.php
```diff
@@ -736,6 +794,11 @@ function _submit_uninstall_reason_action() {
 
 			$this->_storage->store( 'uninstall_reason', $reason );
 
+			if ( ! $this->_plugin->is_plugin() ) {
+				// @todo Check slug/instance.
```
vovafeldman: @fajardoleo not sure what's the @todo means, is it still relevant?

In any case, please add some documentation that explains that with themes there's no uninstallation hook, and we need to trigger it on deactivation.

Also, I would add another helper function to the plugin entity called `is_theme()` and use it instead here. It makes the code clearer.

includes/class-freemius.php
```diff
@@ -5426,6 +5516,10 @@ function is_ssl_and_plan( $plan, $exact = false ) {
 		 * @return string
 		 */
 		function _get_admin_page_url( $page = '', $params = array() ) {
+			if ( ! $this->_plugin->is_plugin() ) {
```
vovafeldman: @fajardoleo Some themes might have settings menu, if that's the case (when there's a menu params set using the dynamic_init), it should work similarly like with plugins. The redirect should go to the settings page, and if the user is in activation / opt-in mode, it should override the default settings (like in plugins) with the opt-in page.

includes/class-fs-plugin-updater.php
```diff
@@ -186,7 +191,12 @@ function pre_set_site_transient_update_plugins_filter( $transient_data ) {
 					$plugin_details->new_version = $new_version->version;
 					$plugin_details->url         = WP_FS__ADDRESS;
 					$plugin_details->package     = $new_version->url;
-					$plugin_details->plugin      = $this->_fs->get_plugin_basename();
+
+					if ( $this->_fs->get_plugin()->is_plugin() ) {
```
vovafeldman: @fajardoleo I suggest to add `is_plugin()` as a helper method in `Freemius` and use it instead to make the code cleaner.

includes/fs-core-functions.php
```diff
@@ -66,13 +66,25 @@ function fs_get_template( $path, &$params = null ) {
 	--------------------------------------------------------------------------------------------*/
 	function fs_enqueue_local_style( $handle, $path, $deps = array(), $ver = false, $media = 'all' ) {
 		global $fs_core_logger;
+
+		$css_path    = fs_normalize_path( WP_FS__DIR_CSS );
```
vovafeldman: @fajardoleo the following 4 lines (including the current one), are repeating in `fs_img_url()` but with `WP_FS__DIR_IMG` instead. You can create a global helper function for code reusability.

And BTW, isn't a similar logic should be added to `fs_enqueue_local_script()`?

includes/fs-essential-functions.php
```diff
@@ -292,15 +292,25 @@ function fs_find_caller_plugin_file() {
 	function fs_update_sdk_newest_version( $sdk_relative_path, $plugin_file = false ) {
 		global $fs_active_plugins;
 
+		$newest_sdk = $fs_active_plugins->plugins[ $sdk_relative_path ];
```
vovafeldman: @fajardoleo About all this function - isn't the fact that you changed risks the whole `start.php` logic? Because when a theme with the new SDK will be activated for the first time it will load the old of this function (`fs_update_sdk_newest_version()`)? 

I didn't check/test it, it just something suspicious that I spotted.

includes/fs-essential-functions.php
```diff
@@ -292,15 +292,25 @@ function fs_find_caller_plugin_file() {
...
+			$in_activation = ( $newest_sdk->plugin_path != $theme->stylesheet );
+		}
+
 		$fs_active_plugins->newest = (object) array(
 			'plugin_path'   => $plugin_file,
 			'sdk_path'      => $sdk_relative_path,
-			'version'       => $fs_active_plugins->plugins[ $sdk_relative_path ]->version,
-			'in_activation' => ! is_plugin_active( $plugin_file ),
+			'version'       => $newest_sdk->version,
+			'fs_for_themes' => ( isset( $newest_sdk->fs_for_themes ) ? $newest_sdk->fs_for_themes : false ),
```
vovafeldman: @fajardoleo Instead of using `'fs_for_themes'` it would be more readable if you'll name the variable `'type'` and it will contain `'theme'` or `'plugin'` as one of the optional values.

includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -360,9 +360,16 @@ function get_top_level_menu_slug() {
...
 		 */
 		function is_activation_page() {
-			return isset( $_GET['page'] ) &&
-			       ( ( strtolower( $this->_menu_slug ) === strtolower( $_GET['page'] ) ) ||
-			         ( strtolower( $this->_plugin_slug ) === strtolower( $_GET['page'] ) ) );
+			$fs = freemius( $this->_plugin_slug );
+
+			if ( $fs->get_plugin()->is_plugin() ) {
+				return isset( $_GET['page'] ) &&
+				       ( ( strtolower( $this->_menu_slug ) === strtolower( $_GET['page'] ) ) ||
```
vovafeldman: @fajardoleo there's a helper function of that (and the following line: `fs_is_plugin_page()`

vovafeldman: And I see it's actually leftovers from my code and not yours :)

includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -360,9 +360,16 @@ function get_top_level_menu_slug() {
...
+				         ( strtolower( $this->_plugin_slug ) === strtolower( $_GET['page'] ) ) );
+			}
+
+			global $pagenow;
```
vovafeldman: @fajardoleo as mentioned in one of the comments above, when there's an admin menu item params set with the init SDK, then this function should use the same logic and return `true` based on the similar logic as with plugin type.

start.php
```diff
@@ -31,11 +31,36 @@
 
 	global $fs_active_plugins;
 
-	$this_sdk_relative_path = plugin_basename( dirname( __FILE__ ) );
+    $paths = array(
```
vovafeldman: I don't see the value of putting the paths in a `$paths` variable, instead of having separated variables.

start.php
```diff
@@ -31,11 +31,36 @@
...
 	global $fs_active_plugins;
 
-	$this_sdk_relative_path = plugin_basename( dirname( __FILE__ ) );
+    $paths = array(
+        'file_path'          => __FILE__,
+        'current_theme_path' => trailingslashit( get_stylesheet_directory() ),
+        'themes_directory'   => trailingslashit( get_theme_root() )
+    );
+
+    if ( function_exists( 'wp_normalize_path' ) ) {
```
vovafeldman: @fajardoleo why not to put the normalization code below the `fs-essential-functions.php` include with `fs_normalize_path()`?

vovafeldman: @fajardoleo I mean after line 63.

start.php
```diff
@@ -53,10 +78,17 @@
 	if ( ! isset( $fs_active_plugins->plugins[ $this_sdk_relative_path ] ) ||
 	     $this_sdk_version != $fs_active_plugins->plugins[ $this_sdk_relative_path ]->version
 	) {
+		if ( $fs_for_themes ) {
```
vovafeldman: @fajardoleo `$fs_for_themes` is not a good name. Instead, use `$is_theme`.

start.php
```diff
@@ -159,7 +211,13 @@
 	}
 
 	if ( version_compare( $this_sdk_version, $fs_active_plugins->newest->version, '<' ) ) {
-		$newest_sdk_starter = fs_normalize_path( WP_PLUGIN_DIR . '/' . $fs_active_plugins->newest->sdk_path . '/start.php' );
+		$newest_sdk = $fs_active_plugins->plugins[ $fs_active_plugins->newest->sdk_path ];
+
+		if ( ! isset( $newest_sdk->fs_for_themes ) || ! $newest_sdk->fs_for_themes ) {
```
vovafeldman: @fajardoleo cleaner version:
`$plugins_or_theme_dir_path = ( ! isset( $newest_sdk->fs_for_themes ) || ! $newest_sdk->fs_for_themes ) ?`
`    WP_PLUGIN_DIR :`
`    get_theme_root();`
`$newest_sdk_starter = fs_normalize_path( $plugins_or_theme_dir_path . '/' . $fs_active_plugins->newest->sdk_path . '/start.php' );`

templates/deactivation-feedback-modal.php
```diff
@@ -42,18 +42,34 @@
...
 		$deactivateLink.click(function (evt) {
 			evt.preventDefault();
 
+			redirectLink = $( this ).attr( 'href' );
+
+			showModal();
+		});
+		<?php
+		} else { ?>
+		$( 'body' ).on( 'click', '.theme-actions .button.activate', function( evt ) {
```
vovafeldman: @fajardoleo missing documentation.

templates/theme-connect.php
```diff
@@ -0,0 +1,197 @@
+<?php
```
vovafeldman: @fajardoleo As mentioned in the original task description, there should only be one connect/opt-in template (unless there are too many differences). So templates/theme-connect.php is redundant and the relevant logic should be in `connect.php`.

templates/theme-icon.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
vovafeldman: @fajardoleo Same about theme-icon.php, it doesn't make sense to have almost the same code twice (unless there are too many differences).

When you adjust the plugin-icon.php, look at issue #67, and adjust the logic based on the suggestion. If it's not clear how to do it, LMK

start.php
fajardoleo: The paths are in the `$paths` variable to avoid duplicating the `if-else` block that normalizes them. The logic is related to the way the `fs-essential-functions.php` file is included. If I put the normalization code after line 63 as you mentioned, the paths will not be normalized in the theme's SDK's starter file if a plugin that uses an older SDK has already set the `$fs_active_plugins` global variable.

It seems that I have to deal with or check the way the `fs-essential-functions.php` file is included first so that I can modify the logic related to the `$paths` variable.

vovafeldman: I don't understand why:

> If I put the normalization code after line 63 as you mentioned, the paths will not be normalized in the theme's SDK's starter file if a plugin that uses an older SDK has already set the $fs_active_plugins global variable.

What if you take all the following code:

    $paths = array(
        'file_path'          => __FILE__,
        'current_theme_path' => trailingslashit( get_stylesheet_directory() ),
        'themes_directory'   => trailingslashit( get_theme_root() )
    );
    if ( function_exists( 'wp_normalize_path' ) ) {
        $paths = array_map( 'wp_normalize_path', $paths );
    } else {
        foreach ( $paths as $type => $path ) {
            $paths[ $type ] = str_replace( '\\', '/', $path );
            $paths[ $type ] = preg_replace( '|/+|', '/', $path );
        }
    }
    $fs_root_path = dirname( $paths[ 'file_path'] );
    if ( false !== strpos( $fs_root_path, $paths['current_theme_path'] ) ) {
	    $this_sdk_relative_path = '../themes/' . str_replace( $paths['themes_directory'], '', $fs_root_path );
        $is_theme = true;
    } else {
        $this_sdk_relative_path = plugin_basename( $fs_root_path );
        $is_theme = false;
    }

And move it below:

	if ( ! function_exists( 'fs_find_direct_caller_plugin_file' ) ) {
		require_once dirname( __FILE__ ) . '/includes/supplements/fs-essential-functions-1.1.7.1.php';
	}

Will it not work?

vovafeldman:
@fajardoleo I just realized that we might have collisions if a plugin & theme both using Freemius and have the same slug. I guess we need to add some hierarchy or maybe a `type` property to the different storage entities.

includes/class-freemius.php
```diff
@@ -227,6 +227,9 @@ class Freemius extends Freemius_Abstract {
 		 */
 		private static $_instances = array();
 
+		// Module types
+		const MODULE_TYPE_PLUGIN = 'plugin';
```
vovafeldman: @fajardoleo I'm not 100% sure, but I think that class `const` is not supported in PHP 5.2. If it does please comment here, so I'll know as well. If it doesn't, please change that to defines.

includes/class-freemius.php
```diff
@@ -264,9 +269,24 @@ private function __construct( $slug ) {
...
+			/**
+			 * @todo Check if there is a better way to set $this->_storage->was_plugin_loaded to true for "theme" module.
+			 * If there is a better way to do that, the following if-else block will not be needed anymore and the original
+			 * code can be brought back.
+			 *
+			 * Since the current code immediately sets $this->_storage->was_plugin_loaded to "true" for "theme" module,
+			 * should_stop_execution() will return "false" and the "few-plugin-tweaks" admin notice can be added with an
+			 * empty title. So the following if-else block ensures that the admin notice will always have a title.
+			 */
+			if ( ! is_object( $this->_plugin ) && $this->is_theme() ) {
```
vovafeldman: @fajardoleo why not to use `$plugin_title = $this->get_plugin_name();` in both cases? (theme & plugin)

includes/class-freemius.php
```diff
@@ -509,23 +547,69 @@ private function _register_account_hooks() {
...
+		}
+
+
+		/**
+		 * Identifies the caller type: plugin or theme.
+		 *
+		 * @author Leo Fajardo (@leorw)
+		 * @since  1.2.0
+		 *
+		 * @return string
```
vovafeldman: @fajardoleo the method doesn't return anything.

includes/class-freemius.php
```diff
@@ -1885,15 +1981,21 @@ private function get_email_sections() {
 		 * @param string $public_key
 		 * @param bool   $is_live
 		 * @param bool   $is_premium
+		 * @param string $type
 		 */
-		function init( $id, $public_key, $is_live = true, $is_premium = true ) {
+		function init( $id, $public_key, $is_live = true, $is_premium = true, $type = Freemius::MODULE_TYPE_PLUGIN ) {
```
vovafeldman: @fajardoleo I think that we agreed that there's no need to have the type in the init due to the automated type identification.

includes/fs-core-functions.php
```diff
@@ -64,15 +64,33 @@ function fs_get_template( $path, &$params = null ) {
 
 	/* Scripts and styles including.
 	--------------------------------------------------------------------------------------------*/
+	function fs_asset_url( $asset_abs_path ) {
```
vovafeldman: @fajardoleo missing documentation. I spend 2 min to understand why is this function needed. My guess is that it's related to the fact that we have themes and plugins now, but I still don't understand why. I would recommend adding an example in the documentation to make it even clearer.

includes/fs-essential-functions.php
vovafeldman: @fajardoleo you didn't respond on this one.

includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -120,7 +126,13 @@ private function get_bool_option( &$options, $key, $default = false ) {
 		 * @param bool  $is_addon
 		 */
 		function init( $menu, $is_addon = false ) {
-			$this->_menu_slug = $menu['slug'];
+			$this->_menu_exists = ( isset( $menu['slug'] ) && ! empty( $menu['slug'] ) );
+
+			if ( $this->_menu_exists ) {
```
vovafeldman: @fajardoleo Turn it to ` = ? : ;` statement.

includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -220,6 +232,16 @@ function has_custom_parent() {
 		}
 
 		/**
+		 * @author Leo Fajardo (@leorw)
+		 * @since  1.2.0
+		 *
+		 * @return bool
+		 */
+		function menu_exists() {
```
vovafeldman: @fajardoleo when it's boolean methods, it's better to call it `is_menu_exists()` or `has_menu()`. Pick your favorite. It's also relevant for the private member - `$this->_menu_exists`

includes/class-freemius.php
fajardoleo: I see. It is not supported in PHP 5.2.

templates/account.php
```diff
@@ -307,8 +307,11 @@ class="dashicons dashicons-cart"></i> <?php ( $show_upgrade ) ?
 									<a target="_blank" class="button button-primary"
 									   href="<?php echo $fs->_get_latest_download_local_url() ?>"><?php echo sprintf( __fs( 'download-x-version', $slug ), $site->plan->title ) . ( is_object( $update ) ? ' [' . $update->version . ']' : '' ) ?></a>
 								<?php elseif ( is_object( $update ) ) : ?>
+									<?php
+										$module_type = $fs->get_module_type();
+									?>
 									<a class="button button-primary"
-									   href="<?php echo wp_nonce_url( self_admin_url( 'update.php?action=upgrade-plugin&plugin=' . $fs->get_plugin_basename() ), 'upgrade-plugin_' . $fs->get_plugin_basename() ) ?>"><?php echo __fs( 'install-update-now', $slug ) . ' [' . $update->version . ']' ?></a>
+									   href="<?php echo wp_nonce_url( self_admin_url( "update.php?action=upgrade-{$module_type}&$module_type=" . $fs->get_plugin_basename() ), "upgrade-{$module_type}_" . $fs->get_plugin_basename() ) ?>"><?php echo __fs( 'install-update-now', $slug ) . ' [' . $update->version . ']' ?></a>
```
vovafeldman: @fajardoleo I think you have a `$` typo in `"update.php?action=upgrade-{$module_type}&$module_type="`. If it's not a typo and you did mean to use the variable, then use `"update.php?action=upgrade-{$module_type}&{$module_type}="` to prevent confusion.

includes/class-freemius.php
fajardoleo: @vovafeldman Sorry, could you please clarify? The method returns the caller module's path and this line in the constructor retrieves and saves the module's path `$this->identify_caller_main_file_and_type();`.

templates/connect.php
```diff
@@ -48,6 +48,16 @@
 
 	$require_license_key = $is_premium_only ||
 	                       ( $is_freemium && $is_premium_code && fs_request_get_bool( 'require_license', true ) );
+
+	global $pagenow;
+	$is_theme_page = ( 'themes.php' === $pagenow );
+?>
+<?php
+if ( $fs->is_theme() && $is_theme_page ) { ?>
```
vovafeldman: @fajardoleo do NOT show the close button if `$fs->is_premium()`

templates/connect.php
```diff
@@ -178,7 +188,7 @@ class="button button-secondary" tabindex="2"><?php _efs( 'skip', $slug ) ?></a>
 			),
 			'events'  => array(
 				'icon-class' => 'dashicons dashicons-admin-plugins',
-				'label'      => __fs( 'permissions-events' ),
+				'label'      => sprintf( __fs( 'permissions-events' ), ucfirst( $fs->get_module_type() ) ),
```
vovafeldman: @fajardoleo missing `, $slug` in the `_fs()` call.

templates/connect.php
```diff
@@ -325,9 +357,18 @@ class="fs-permission fs-<?php esc_attr_e( $id ); ?>">
 		 * @author Vova Feldman (@svovaf)
 		 * @since 1.1.9
 		 */
-		var $connectLicenseModeTrigger = $('#fs_connect .fs-freemium-licensing a');
+		var
+			$connectLicenseModeTrigger = $('#fs_connect .fs-freemium-licensing a'),
+			href = window.location.href;
+
+		if ( -1 === href.indexOf( '?' ) ) {
```
vovafeldman: @fajardoleo I think that `indexOf` return different values in different browsers. I would instead use `if ( 0 < href.indexOf( '?' ) )` and switch the logic between the if and else.

templates/deactivation-feedback-modal.php
```diff
@@ -40,20 +40,42 @@
 		    + '		</div>'
 		    + '	</div>'
 		    + '</div>',
-	    $modal = $(modalHtml),
-	    $deactivateLink = $('#the-list .deactivate > [data-slug=<?php echo $VARS['slug']; ?>].fs-slug').prev(),
-	    selectedReasonID = false;
+	    $modal           = $(modalHtml),
+	    $deactivateLink  = $('#the-list .deactivate > [data-slug=<?php echo $VARS['slug']; ?>].fs-slug').prev(),
+		selectedReasonID = false,
+		pluginSlug       = '<?php echo $slug; ?>',
```
vovafeldman: @fajardoleo since it's a new var, I would already call it `moduleSlug`.

templates/plugin-icon.php
```diff
@@ -49,24 +55,54 @@
...
+						'fields' => $fields
+					) );
+				}
+
+				if ( ! is_wp_error( $plugin_or_theme_information ) ) {
+					// Not sure if "icons" or "screenshots" will always be set.
+					if ( isset( $plugin_or_theme_information->icons ) && ! empty( $plugin_or_theme_information->icons ) ) {
+						// Get the smallest icon.
+						$icon = end( $plugin_or_theme_information->icons );
+					} else if ( isset( $plugin_or_theme_information->screenshots ) && ! empty( $plugin_or_theme_information->screenshots ) ) {
```
vovafeldman: @fajardoleo if the icon is not set, do NOT use the screenshots as a fallback.

includes/fs-essential-functions.php
fajardoleo: @vovafeldman Sorry, I am not sure if I understand the issue. Could you please clarify? When the old `fs_update_sdk_newest_version` is called by a theme, `is_plugin_active` will always return `false` so `in_activation` will always be `true`. When this happens, the old SDK will be able to load/start the new SDK since the value of `$is_newest_sdk_path_valid` in `start.php` of the old SDK will be `true`.

includes/class-freemius.php
fajardoleo: @vovafeldman Because `get_plugin_name()` will do more work if `$this->_plugin_name` is not yet set. So if `$this->_plugin` is already set and `$this->_plugin_name` is not yet set, that "more work" can be avoided if I will just use `$this->_plugin`. I may need to dig deeper to understand why this behavior only happens with plugins (`$this->_plugin` already set and `$this->_plugin_name` not yet set).

vovafeldman: @fajardoleo then add a note about that.

includes/class-freemius.php
vovafeldman: @fajardoleo the method calculates that, but it doesn't actually return anything. So `@return string` should be removed from the function documentation.

fajardoleo: Ok, I see now. I was confused by the diff. I only looked at the `_find_caller_plugin_file` method and didn't notice the `@return string` at the bottom of the diff :)

includes/class-freemius.php
fajardoleo: OK, code updated.

includes/managers/class-fs-admin-menu-manager.php
fajardoleo: @vovafeldman Code updated.

includes/managers/class-fs-admin-menu-manager.php
fajardoleo: @vovafeldman Code updated.

templates/plugin-icon.php
fajardoleo: @vovafeldman `$plugin_or_theme_information->icons` is only for plugins and `$plugin_or_theme_information->screenshots` is only for themes.

templates/deactivation-feedback-modal.php
fajardoleo: @vovafeldman Code updated.

templates/plugin-icon.php
vovafeldman: @fajardoleo right! Forgot that themes doesn't have icons. I will ask Stanley to design a simple default icon for themes. No need to use screenshot anyways. For now, create a copy of `plugin-icon.png` to `theme-icon.png`, and use that for themes. No need to ping the API, and no need to create a copy of the file. Just use the `theme-icon.png` when it's a theme.

includes/fs-essential-functions.php
vovafeldman: @fajardoleo just wanted to make sure you tested that part.

fajardoleo: @vovafeldman Yes, I tested that part.

includes/class-freemius.php
fajardoleo: @vovafeldman Code updated.

includes/fs-core-functions.php
fajardoleo: @vovafeldman Code updated.

includes/class-freemius.php
fajardoleo: @vovafeldman Code updated.

templates/connect.php
fajardoleo: @vovafeldman Code updated. I used `if ( href.indexOf( '?' ) > 0 )` instead of `if ( 0 < href.indexOf( '?' ) )` since the latter is hard to read and WP's Coding Standards page also mentions:
>Yoda conditions for <, >, <= or >= are significantly more difficult to read and are best avoided.

templates/account.php
fajardoleo: @vovafeldman Code updated.

templates/connect.php
fajardoleo: @vovafeldman Code updated.

templates/connect.php
fajardoleo: @vovafeldman Code updated.

templates/plugin-icon.php
fajardoleo: @vovafeldman Code updated.

includes/class-freemius.php
fajardoleo: @vovafeldman Sorry, please ignore my previous comment. Class constants are actually supported in 5.0: http://php.net/manual/en/migration51.oop.php#migration51.oop-constants.

PR #75, opened 2016-07-29
Feature/unregister uninstall hook

fajardoleo:


vovafeldman:
Here's the new fixed flow:
1. Free version is activated
2. Free version deactivated
3. Free version registers free version uninstall hook
4. Premium version activated
5. Premium version activation hook will clear all plugin's uninstall hooks
6. If the user uninstalls the free version now, no uninstall hooks are registered

PR #80, opened 2016-08-11
[deactivation-form] [reasons] Randomize the deactivation reasons to m…

fajardoleo:
…ake sure that the feedback given by users is authentic (e.g.: a user may choose the first reason because it's just quicker).

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -708,9 +708,6 @@
 				$reason_dont_share_info['internal_message'] = fs_get_template( 'forms/deactivation/retry-skip.php', $internal_message_template_var );
 			}
 
-			$long_term_user_reasons[] = $reason_temporary_deactivation;
-			$long_term_user_reasons[] = $reason_other;
-
 			$uninstall_reasons = array(
 				'long-term'                                   => $long_term_user_reasons,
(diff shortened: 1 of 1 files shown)


danieliser:
In the case they choose the first answer for convenience, wouldn't this actually make that harder to detect? IE they will still choose the first answer, but now it will be random making it impossible to detect.

Instead why not put a empty answer in the first spot, and either mark the field required or let it submit blank. A blank answer would at least be trackable as not answered. Randomizing it means no possible way to detect pattern, plus adds false positives to the existing reasons.

vovafeldman:
@danieliser we don't know how much the positioning affects the selection, but it probably does. 
We do see that some plugins' top uninstallation reason is not the 1st bullet, but for most it is. 
So we want to eliminate the order affect on the results. 

By randomizing the order, if a subset of the users is just picking the 1st option, it will be equally distributed, and will help devs to identify the main uninstallation reasons by removing the noise.

danieliser:
@vovafeldman Though that may be the goal, randomly distributing will actually skew the results. Why not just make the first option "Don't Feel Like Taking Time To Tell You". Would accomplish the same goal no? I would argue I prefer to not have results for the other choices flooded with inaccurate results.

PR #83, opened 2016-08-18
Feature/deactivation form optin

vovafeldman:


includes/class-freemius.php
```diff
@@ -781,9 +797,13 @@ function _submit_uninstall_reason_action() {
 
 			$reason_info = isset( $_REQUEST['reason_info'] ) ? trim( stripslashes( $_REQUEST['reason_info'] ) ) : '';
 
+			// Check if feedback should be sent anonymously.
+			$anonymous = ( isset( $_REQUEST['anonymous'] ) && 'true' == $_REQUEST['anonymous'] );
```
vovafeldman: @fajardoleo you should use `fs_request_get_bool('anonymous');` instead.

includes/class-freemius.php
```diff
@@ -781,9 +797,13 @@ function _submit_uninstall_reason_action() {
...
+
 			$reason = (object) array(
-				'id'   => $_POST['reason_id'],
-				'info' => substr( $reason_info, 0, 128 )
+				'id'        => $_POST['reason_id'],
```
vovafeldman: @fajardoleo you have to validate that the parameter is unsigned integer. In general, you should never use user given params without validating them.

includes/class-freemius.php
vovafeldman: @fajardoleo I would also refactor the name of the var to `$is_anonymous`

PR #84, opened 2016-08-23
Feature/menu permission

fajardoleo:


includes/class-freemius.php
```diff
@@ -6758,6 +6758,29 @@ private function override_plugin_menu_with_activation() {
...
+			global $menu;
+
+			foreach ( $menu as $menu_info ) {
+				/**
+				 * The second element in the menu info array is the capability/role that has access to the menu and the
+				 * third element is the menu slug.
+				 */
+				if ( $menu_info[2] === $this->get_top_level_menu_slug() ) {
+					return $menu_info[1];
+					break;
```
vovafeldman: @fajardoleo there's no need for `break` after `return` statement.

includes/class-freemius.php
```diff
@@ -6876,7 +6901,7 @@ private function embed_submenu_items() {
 								null,
 							$item['page_title'],
 							sprintf( $item_template, $item['menu_title'] ),
-							$item['capability'],
+							$item['show_submenu'] ? $top_level_menu_capability : $item['capability'],
```
vovafeldman: @fajardoleo when the submenu is visible (`$item['show_submenu'] == true`) and the top level menu capability will be `'read'`, it means that pages like **Account** will be visible for all admin users. And this is not the expected functionality. I think that this line should be changed to:
`!empty($item['capability']) ? $item['capability'] : $top_level_menu_capability`

And the same code should be set on line `6917`.

And then, change the capability param at:
```
					$this->add_submenu_link_item(
						$this->apply_filters( 'support_forum_submenu', __fs( 'support-forum', $this->_slug ) ),
						$this->apply_filters( 'support_forum_url', 'https://wordpress.org/support/plugin/' . $this->_slug ),
						'wp-support-forum',
						'read',
						50
					);
```

from `'read'` to `null` so it will inherit the capability from the parent.

includes/class-freemius.php
fajardoleo: Didn't notice that :smile:

includes/class-freemius.php
fajardoleo: @vovafeldman Changes have been pushed.

PR #91, opened 2016-09-14
added functions for free version deactivation

bradvin:
added function to check if free version activated
added function to deactivate free version


vovafeldman:
We already added the logic, will be added on the next version.

PR #94, opened 2016-09-26
Feature/fs theme mode v2

fajardoleo:


includes/class-freemius.php
```diff
@@ -570,62 +633,116 @@ private function _register_account_hooks() {
...
+		 *
+		 * @since 1.2.1
 		 */
-		private function _find_caller_plugin_file( $is_init = false ) {
-			// Try to load the cached value of the file path.
-			if ( isset( $this->_storage->plugin_main_file ) ) {
-				if ( file_exists( $this->_storage->plugin_main_file->path ) ) {
-					return $this->_storage->plugin_main_file->path;
-				}
+		function upgrade( $module_id, $slug ) {
```
fajardoleo: @vovafeldman Yes, it should actually be private.

includes/class-freemius.php
```diff
@@ -4595,13 +4763,49 @@ private static function get_all_sites() {
...
+		 * @author Leo Fajardo (@leorw)
+		 *
+		 * @since 1.2.1
+		 *
+		 * @param string $option_name
+		 * @param mixed  $option_value
+		 * @param bool   $store
+		 * @param number $module_id
+		 */
+		private function set_account_option( $option_name, $option_value, $store ) {
```
fajardoleo: @vovafeldman `get_account_option()` is only called in static methods. Two of those caller methods (`get_all_sites` and `get_all_licenses`) are called in `_debug_page_render` method that doesn't know about the module ID. If I will change `get_account_option()` to a non-static method, then I guess I will also have to change the mentioned caller methods so that they will receive the module ID instead of the module type? Then I can get the instance from the module ID. After that, I will also have to modify `_debug_page_render` and get it to load all sites and licenses using a new method or using `self::$_accounts` directly.

vovafeldman: If it's **only** called in static methods then no need.

includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -83,22 +94,23 @@ class FS_Admin_Menu_Manager {
...
 		}
 
-		protected function __construct( $plugin_slug ) {
-			$this->_logger = FS_Logger::get_logger( WP_FS__SLUG . '_' . $plugin_slug . '_admin_menu', WP_FS__DEBUG_SDK, WP_FS__ECHO_DEBUG_SDK );
+		protected function __construct( $module_id ) {
+			$this->_logger = FS_Logger::get_logger( WP_FS__SLUG . '_' . $module_id . '_admin_menu', WP_FS__DEBUG_SDK, WP_FS__ECHO_DEBUG_SDK );
 
-			$this->_plugin_slug = $plugin_slug;
+			$this->_module_id   = $module_id;
+			$this->_module_unique_affix = freemius( $module_id )->get_unique_affix( '-' );
```
fajardoleo: @vovafeldman May I know if that means that I should also avoid using `Freemius` or `freemius( $module_id )` inside `fs_get_unique_affix()`?

includes/class-freemius.php
```diff
@@ -308,6 +355,18 @@ private function __construct( $slug, $is_init = false ) {
...
 
 		/**
+		 * Checks whether this plugin or theme has settings menu.
+		 *
+		 * @author Leo Fajardo (@leorw)
+		 * @since  1.2.0
+		 *
+		 * @return bool
+		 */
+		private function has_settings_menu() {
```
fajardoleo: @vovafeldman It is now used by themes.

includes/class-freemius.php
```diff
@@ -2717,7 +2710,7 @@ function is_addon_activated( $addon_id ) {
 		 * @return bool
 		 */
 		function is_addon_connected( $addon_id ) {
-			$sites = self::get_all_sites( Freemius::MODULE_TYPE_PLUGIN );
+			$sites = self::get_all_sites( $this->_module_type );
```
vovafeldman: @fajardoleo In the current case, since add-ons are always plugins, it should be `Freemius::MODULE_TYPE_PLUGIN`.

config.php
```diff
@@ -175,6 +175,14 @@
 	                                    strtoupper( $_SERVER['REQUEST_METHOD'] ) == 'POST' ) );
 
 	/**
+	 * Module types
+	 *
+	 * @since 1.2.2
+	 */
+	define( 'MODULE_TYPE_PLUGIN', 'plugin' );
```
vovafeldman: @fajardoleo please refactor the define names to have `WP_FS__` prefix.

includes/class-freemius.php
```diff
@@ -253,19 +266,33 @@ class Freemius extends Freemius_Abstract {
...
+		 * @param bool   $is_init Since 1.2.2 Is initiation sequence.
 		 */
-		private function __construct( $slug, $is_init = false ) {
-			$this->_slug = $slug;
+		private function __construct( $module_id, $slug = false, $is_init = false ) {
+			if ( ! is_numeric( $module_id ) ) {
+				$slug = $module_id;
 
-			$this->_logger = FS_Logger::get_logger( WP_FS__SLUG . '_' . $slug, WP_FS__DEBUG_SDK, WP_FS__ECHO_DEBUG_SDK );
+				$this->_plugin = FS_Plugin_Manager::instance( $slug, $slug, WP_FS__MODULE_TYPE_PLUGIN )->get();
```
vovafeldman: @fajardoleo I'm not sure why you שרק loading the plugin's data here if you already loading it later in the constructor
`$this->_plugin = FS_Plugin_Manager::instance( $this->_module_id, $this->_slug, $this->_module_type )->get();`

includes/class-freemius.php
```diff
@@ -253,19 +266,33 @@ class Freemius extends Freemius_Abstract {
 		 * @author Vova Feldman (@svovaf)
 		 * @since  1.0.0
 		 *
+		 * @param number $module_id
 		 * @param string $slug
-		 * @param bool   $is_init Since 1.2.1 Is initiation sequence.
+		 * @param bool   $is_init Since 1.2.2 Is initiation sequence.
```
vovafeldman: @fajardoleo the `$is_init` was introduced in 1.2.1

includes/class-freemius.php
```diff
@@ -617,15 +677,104 @@ private function _find_caller_plugin_file( $is_init = false ) {
...
+
+		/**
+		 * @author Leo Fajardo (@leorw)
+		 * @since  1.2.2
+		 *
+		 * @param  number $module_id
+		 *
+		 * @return array
+		 */
+		static function get_slug_and_type_info( $module_id ) {
```
vovafeldman: @fajardoleo it looks like this static method isn't used. Please verify.

vovafeldman:
@fajardoleo I reviewed the `$is_init` related code, and it looks fine.

Another question that I have related to the following scenario:
1. Activated pluginX and pluginY when Debug Bar is on.
2. When pluginX was activated, `id_slug_type_path_map` got the right path.
2. Deactivated pluginX. During deactivation, `id_slug_type_path_map` loose the path in `clear_module_main_file_cache()`.

**Do you see any issue(s) when `id_slug_type_path_map` don't have the `path` property when the module is deactivated and `freemius($pluginX_slug)` is called for example from the Freemius debug page?**

PR #100, opened 2016-10-24
Fix some issues raised by Theme Check plugin

primozcigler:
Hi,

I found time to fix some issues that Theme Check plugin raises when Freemius SDK is included in the theme. It is only the tip of the iceberg, but it is a start.

## Commit ec6766a

Usually `WP_DEBUG` constant can be used to achieve full reporting. Maybe I missed something?

## Commit 349897f

`esc_attr_e` and `esc_html_e` are i18n functions merged with sanitization functions. If you're printing out a variable, the respective `esc_attr` and `esc_html` should be used (no `_e` at the end).

## Commit ff2a340

ASCII characters are safe in files. I changed the bullet char with the HTML entity `&bull;`.

---

Sharing my view on the rest of the errors:

- `WARNING: Found a translation function that is missing a text-domain.`

  You will need to go through each of these and add text-domains.

- `WARNING: Found base64_encode in the file ...`

  @vovafeldman mentioned the hack in the email, I left these untouched.

- `WARNING: fopen/fclose/curl_* was found in the file ...`

  It will take me too much time to reverse engineer this code. It is best that you refactor this code, using `WP_Filesystem` and/or HTTP API, so you don't need to use these functions directly anymore.

- `REQUIRED: ... Themes should use add_theme_page() for adding admin pages.`

  All instances of `add_menu_page`, `add_submenu_page` and `add_debug_page` should be replaced with `add_theme_page`. If you need some these pages for testing/debugging, I would recommend that you put somewhere in docs which files can be removed during the preparation of the theme for upload, so the code triggering errors can be removed upon testing.

After that there are lots of INFO notices, which AFAIK are not required to follow, but I would strive to fix as many of these as well.

Primoz

### DIFF
--- a/config.php
+++ b/config.php
@@ -226,9 +226,6 @@
 
 	if ( WP_FS__ECHO_DEBUG_SDK ) {
 		error_reporting( E_ALL );
-		ini_set( 'error_reporting', E_ALL );
-		ini_set( 'display_errors', true );
-		ini_set( 'html_errors', true );
 	}
 
 
--- a/includes/fs-plugin-info-dialog.php
+++ b/includes/fs-plugin-info-dialog.php
@@ -597,7 +597,7 @@
 											}
 											?>
 											<a class="nav-tab" data-billing-cycle="<?php echo $cycle ?>"
-											   data-pricing="<?php esc_attr_e( json_encode( $prices ) ) ?>">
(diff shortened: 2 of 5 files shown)


vovafeldman:
Thanks @primozcigler! We'll take care of the rest.

PR #103, opened 2016-11-08
Ignore single-file-plugins when finding the SDK location

bfintal:
The SDK path is incorrectly detected if there are any single-file-plugins that are listed earlier in the plugin list. Ignore single-file-plugins to fix this.


vovafeldman:
Great bug catch! Thanks @bfintal.

PR #106, opened 2016-11-14
Feature/opt out and opt in

fajardoleo:


includes/class-freemius.php
```diff
@@ -4335,8 +4390,9 @@ function _uninstall_plugin_event( $check_user = true ) {
 			} else {
 				// Send uninstall event.
 				$this->send_install_update( array_merge( $params, array(
-					'is_active'      => false,
-					'is_uninstalled' => true,
+					'is_active'       => false,
+					'is_disconnected' => true,
```
vovafeldman: @fajardoleo don't change the `is_disconnected` flag on uninstall, we do want to preserve it.

includes/class-freemius.php
```diff
@@ -5481,6 +5563,16 @@ function _require_license_activation_dialog() {
 		}
 
 		/**
+		 * Includes all needed code for the opt-out dialog.
+		 *
+		 * @author Leo Fajardo (@leorw)
+		 * @since  1.2.2
+		 */
+		function _require_optout_dialog() {
```
vovafeldman: @fajardoleo it seems like it's only used internally in the class so if I'm right, turn it to private and remove the `_` prefix.

includes/class-freemius.php
```diff
@@ -6387,6 +6479,23 @@ function opt_in(
 		) {
 			$this->_logger->entrance();
 
+			if ( ! $is_uninstall && fs_request_is_action( $this->_slug . '_opt_in' ) ) {
```
vovafeldman: @fajardoleo this logic should definitely be in a separate method, probably name it `reconnect()`. And for consistency, I would name the `opt_in()` to `disconnect()`.

includes/class-freemius.php
```diff
@@ -6387,6 +6479,23 @@ function opt_in(
...
+					'is_disconnected' => false
+				) );
+
+				if ( ! $this->is_api_error( $site ) ) {
+					$this->_site = new FS_Site( $site );
+					$this->_store_site();
+
+					$this->schedule_sync_cron();
+
+					fs_redirect( $this->get_after_activation_url( 'after_connect_url' ) );
```
vovafeldman: @fajardoleo don't redirect after reconnecting, instead, add an admin notice on the same page. You can use the same string from the `opt-out-message-appreciation` message.

PR #139, opened 2017-01-25
[new-wp-checkout] Initial commit.

fajardoleo:


templates/checkout.php
```diff
@@ -25,11 +25,18 @@
...
 
 	$context_params = array(
-		'plugin_id'         => $fs->get_id(),
-		'plugin_public_key' => $fs->get_public_key(),
-		'plugin_version'    => $fs->get_plugin_version(),
+		'plugin_id'      => $fs->get_id(),
+		'plugin_version' => $fs->get_plugin_version(),
+        'public_key'     => $fs->get_public_key(),
+        'mode'           => 'dashboard',
+        'plan_id'        => $_GET['plan_id'],
```
vovafeldman: @fajardoleo using `$_GET` params as is without any validation isn't a good practice and might be insecure in some cases, please verify the param type with `FS_Plugin_Plan::is_valid_id()` and if it isn't, we should probably NOT add that param to the `$context_params` and let the checkout to auto choose the default plan (the one with `is_featured` `true`).

templates/checkout.php
```diff
@@ -25,11 +25,18 @@
...
+        'trial'          => ( isset( $_GET['trial'] ) ? $_GET['trial'] : false ),
```
vovafeldman: @fajardoleo same thing here, you have to validate that the `trial` is boolean. If I'm not mistaken, you actually need to check if it's a string and one of the following `true` or `false`, because there's no real way to pass booleans via querystring request. So you need to use `in_array( $_GET['trial'], array( 'true', 'false' ) )`.

templates/checkout.php
```diff
@@ -25,11 +25,18 @@
...
 	);
 
+	if ( isset( $_GET['licenses'] ) ) {
+	    $context_params['licenses'] = $_GET['licenses'];
```
vovafeldman: @fajardoleo verify the type (should be int), and if not valid, let the checkout to choose the right one.

PR #144, opened 2017-02-15
Register site from code

gfirem:
Adding a function to register site in the saas to an existing user and license. I need to use inside a multi-site, where a client buy and unlimited license and i dont want to register each site to the same user, i add this function and combine with the hooks:
fs_after_account_connection_
fs_before_admin_menu_init_
fs_after_account_delete_
Please if exist a better way to accomplish, let me know i want to improve this.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -6909,6 +6909,125 @@
 
 			return $decoded;
 		}
+		
+		/**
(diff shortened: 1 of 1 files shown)


vovafeldman:
@gfirem I'm closing this one, we already have the `opt_in()` method that does that.

PR #167, opened 2017-06-19
NPM Module Cleanup

jhabdas:
Broken into three commits, explained here:

- d79c55c empties `dependencies` object as this module has no hard dependencies
- 6dad57c adds an NPM script to prune unnecessary files (based on docs and some ad libbing)
- bf9439f sets NPM module as private, adjusts version and simplifies the package manifest

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,18 +3,6 @@
   "version": "1.0.0",
   "description": "",
   "main": "gulpfile.js",
-  "dependencies": {
-    "gulp": "^3.9.1",
-    "gulp-clean": "^0.3.2",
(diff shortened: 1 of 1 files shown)


vovafeldman:
@jhabdas quick question - is the version of the package should represent the version of the library? Or is it the version of the package?

vovafeldman:
@jhabdas we are preparing for the next release and I'm going through your suggested changes, can you please explain why you set the `private` flag to `true`?

vovafeldman:
Closing due to unresponsiveness from the committer.

PR #171, opened 2017-07-28
Feature/affiliation submenu

fajardoleo:


includes/class-freemius.php
```diff
@@ -6450,6 +6480,196 @@ function is_themes_page() {
...
+            }
+
+		    return $this->_plugin->has_affiliation();
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 1.2.1.7.2
+         */
+        private function load_affiliate_and_terms() {
```
vovafeldman: @fajardoleo for consistency, change it to `fetch_affiliate_and_terms()`. `load_` is when things are loaded from the storage while `fetch_` is when the data fetched from an API.

includes/entities/class-fs-plugin.php
```diff
@@ -59,6 +59,14 @@ class FS_Plugin extends FS_Scope_Entity {
...
 		 */
 		public $is_live;
+		/**
+         * @author Leo Fajardo (@leorw)
+         *
+		 * @since 1.2.1.7.2
+		 *
+		 * @var string
+		 */
+		public $affiliation;
```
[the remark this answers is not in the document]
vovafeldman: @fajardoleo from what I've seen, this value is set by the init SDK params, and the 'affiliation' value is actually boolean. If that's the case, I suggest to rename it to `$has_affiliate_program` (also respectively modify the init_sdk param to `has_affiliate_program` instead of `has_affiliation` and the getter method in this class).

includes/class-freemius.php
fajardoleo: 👍

includes/class-freemius.php
```diff
@@ -6450,6 +6480,196 @@ function is_themes_page() {
...
+		#region Affiliation
+		#----------------------------------------------------------------------------------
+
+        /**
+         * @author Leo Fajardo
+         * @since 1.2.1.7.2
+         *
+         * @return bool
+         */
+        function has_affiliation() {
```
vovafeldman: @fajardoleo refactor to `has_affiliate_program()`

includes/entities/class-fs-affiliate-terms.php
```diff
@@ -0,0 +1,128 @@
...
+            return ( 0 == $this->cookie_days );
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         *
+         * @return bool
+         */
+        function has_renewals_commission() {
+            return is_null( $this->commission_renewals_days );
```
vovafeldman: @fajardoleo if `$this->commission_renewals_days` is a number, it means that there's a renewals commission which is limited to `$this->commission_renewals_days` - days. So semantically, this method name is either wrong or the implementation is wrong. If you want to check if there are unlimited renewals commission, then the method should be `has_unlimited_renewals_commission()`

vovafeldman: P.S. if that's the same on our backend, it should also be updated there

includes/entities/class-fs-plugin.php
```diff
@@ -59,6 +59,17 @@ class FS_Plugin extends FS_Scope_Entity {
 		 * @var bool
 		 */
 		public $is_live;
+		/**
+         * `false` if the module doesn't have an affiliate program or one of the following: `selected`, `customers`,
```
vovafeldman: @fajardoleo the proper place to put the description of the property is after its type (i.e. after `@var string|bool`). I also think that you can specify `string|false` as the type to be more accurate (make sure the IDE doesn't show a warning for it).

PR #181, opened 2017-08-31
[themes] [tabs] [monetization] [add-ons] Only add tabs globally to mo…

vovafeldman:
…netizing themes.


vovafeldman:
forgot to bump the version

PR #190, opened 2017-10-19
[admin-notice] Trigger an affiliate program–related admin notice afte…

fajardoleo:
…r 30 days letting users/customers know about it.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -768,6 +768,7 @@
 			$this->add_filter( 'after_code_type_change', array( &$this, '_after_code_type_change' ) );
 
 			add_action( 'admin_init', array( &$this, '_add_trial_notice' ) );
+			add_action( 'admin_init', array( &$this, '_add_affiliate_program_notice' ) );
 			add_action( 'admin_init', array( &$this, '_enqueue_common_css' ) );
 
 			/**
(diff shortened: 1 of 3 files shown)


includes/class-freemius.php
```diff
@@ -12920,6 +12925,92 @@ function _add_trial_notice() {
...
+             * Do not show the notice if it was already shown before or less than 30 days have passed since the initial
+             * activation with FS.
+             */
+            if ( $was_notice_shown_before ||
+                $this->_storage->install_timestamp > ( time() - ( WP_FS__TIME_24_HOURS_IN_SEC * 30 ) )
+            ) {
+                return false;
+            }
+
+            if (
```
vovafeldman: @fajardoleo it's semantically better to move this logic before checking if the promotional notice was already visible before. For example, if the module doesn't have any affiliation program, then there's no reason to check the timing.

includes/class-freemius.php
```diff
@@ -12920,6 +12925,92 @@ function _add_trial_notice() {
...
+				return false;
+			}
+
+			if ( ! $this->is_paying() &&
+                FS_Plugin::AFFILIATE_MODERATION_CUSTOMERS == $this->_plugin->affiliate_moderation ) {
+			    // If the user is not a customer and the affiliate program is only for customers, don't show the notice.
+                return false;
+			}
+
+			if ( ! $this->apply_filters( 'show_affiliate_program_notice', true ) ) {
```
vovafeldman: @fajardoleo same about this, if the developer specified not to show the notice, it should be checked before the logic that checks if the notice was visible before.

includes/class-freemius.php
```diff
@@ -12920,6 +12925,92 @@ function _add_trial_notice() {
...
+				// Developer explicitly asked not to show the notice about the affiliate program.
+				return false;
+			}
+
+			if ( $this->is_activation_mode() || $this->is_pending_activation() ) {
```
vovafeldman: @fajardoleo same about here.

PR #198, opened 2017-11-06
[paths] Store relative paths instead of absolute paths to the storage.

fajardoleo:


includes/class-freemius.php
```diff
@@ -918,9 +958,49 @@ private function _find_caller_plugin_file( $is_init = false ) {
...
+		    $is_plugin = empty( $module_type ) ?
+                $this->is_plugin() :
+                ( WP_FS__MODULE_TYPE_PLUGIN === $module_type );
+
+		    $module_root_dir = fs_normalize_path( $is_plugin ?
+                WP_PLUGIN_DIR :
+                get_theme_root() );
+
+		    if ( 0 !== strpos( $path, $module_root_dir ) ) {
+		        $path = ( $module_root_dir . '/' . $path );
```
vovafeldman: @fajardoleo I'd recommend using `$path = trailingslashit( $module_root_dir ) . $path;` instead to avoid situations where some environments may already have a trailing slash at the end of the plugin/theme path variable.

includes/class-freemius.php
```diff
@@ -894,8 +933,9 @@ private function _find_caller_plugin_file( $is_init = false ) {
 				if ( isset( $this->_storage->plugin_main_file ) &&
 				     isset( $this->_storage->plugin_main_file->prev_path )
 				) {
-					if ( file_exists( $this->_storage->plugin_main_file->prev_path ) ) {
-						return $this->_storage->plugin_main_file->prev_path;
+                    $absolute_path = $this->get_absolute_path( $this->_storage->plugin_main_file->prev_path );
```
vovafeldman: @fajardoleo I noticed that you've forgot to migrate `$this->_storage->plugin_main_file->prev_path` to the relative path structure (in the migration logic, not here).

includes/class-freemius.php
```diff
@@ -1042,7 +1125,10 @@ private function get_caller_main_file_and_type() {
 
 					if ( $caller_file_path == fs_normalize_path( realpath( trailingslashit( $themes_dir ) . basename( dirname( $caller_file_path ) ) . '/' . basename( $caller_file_path ) ) ) ) {
 						$module_type           = WP_FS__MODULE_TYPE_THEME;
-						$caller_file_candidate = $caller_file_path;
+						$caller_file_candidate = basename( dirname( $caller_file_path ) ) .
```
vovafeldman: @fajardoleo I'll add a comment that this calculates the relative path of the theme + also add an inline example in the comment.

includes/class-freemius.php
```diff
@@ -5855,14 +5941,20 @@ function get_module_label( $lowercase = false ) {
 		 * @author Vova Feldman (@svovaf)
 		 * @since  1.0.4
 		 *
+         * @param string $plugin_main_file_path
+         *
 		 * @return string
 		 */
-		function get_plugin_basename() {
+		function get_plugin_basename( $plugin_main_file_path = false ) {
```
vovafeldman: @fajardoleo you've added `$plugin_main_file_path` as an argument, but I don't see any call to this method with an argument.

includes/class-freemius.php
```diff
@@ -918,9 +958,49 @@ private function _find_caller_plugin_file( $is_init = false ) {
...
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 1.2.2.11
+         *
+         * @param string $path
+         *
+         * @return string
+         */
+		private function get_relative_path( $path ) {
```
vovafeldman: @fajardoleo this method is confusing. It gets a `$path` argument but it doesn't really work for any path beside the path of the current context module.

includes/class-freemius.php
```diff
@@ -918,9 +960,54 @@ private function _find_caller_plugin_file( $is_init = false ) {
...
+         * @param string|bool $module_type
+         *
+         * @return string
+         */
+        private function get_absolute_path( $path, $module_type = false ) {
+            $is_plugin = empty( $module_type ) ?
+                $this->is_plugin() :
+                ( WP_FS__MODULE_TYPE_PLUGIN === $module_type );
+
+            $module_root_dir = fs_normalize_path( $is_plugin ?
```
vovafeldman: @fajardoleo I suggest creating a helper method to calculate the `get_module_root_dir_path( $module_type = false )` since you have it here and in `get_relative_path()`.

PR #199, opened 2017-11-08
Fix to class-freemius.php displaying 1 instead of the literal plugin

alanef:
Code inadvertently converts 'true' to lowercase rather than the literal string, hence displaying 1 - as per issue raised on  Slack


vovafeldman:
Thanks for the PR @alanef. Apparently, it was already fixed a while ago in our `develop` branch and will be released with the upcoming version.

PR #206, opened 2017-12-08
Fix get_module_label function to return correct label 💯

asharirfan:
Hi guys,

`get_module_label` function, in `class-freemius.php`, was returning `string 1` because the label in the function was getting replaced by `$lowercase` variable.

I have corrected the mistake and sending a pull request.

### Screenshot

alt text

Cheers!

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -2259,7 +2259,7 @@
 
 			return ( checkdnsrr( $domain, 'MX' ) || checkdnsrr( $domain, 'A' ) );
 		}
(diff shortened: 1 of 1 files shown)


vovafeldman:
Thanks for the PR @asharirfan, we already pushed a fix into the `develop` branch on September (https://github.com/Freemius/wordpress-sdk/commit/3310bf6bb464e299019d38e98c784fe299aa664a). We are planning to release the new SDK very soon so you'll be covered :+1:

PR #232, opened 2018-05-13
Feature/eu gdpr admin notice

fajardoleo:


includes/class-freemius.php
```diff
@@ -629,6 +636,254 @@ private function _version_updates_handler() {
...
+            ) {
+                add_action( 'init', array( &$this, '_maybe_show_gdpr_admin_notice' ) );
+            }
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since  2.1.0
+         */
+        function _maybe_show_gdpr_admin_notice() {
```
vovafeldman: @fajardoleo since this method is executed per product, it involves API calls which can take time. So we can get into a theoretical situation where an admin opens the admin page several times in parallel right after an update of the SDK, which can run this logic several times. Unfortunately, PHP doesn't have locks/mutexes. BUT, we can significantly reduce the chance of that happening by simulating a lock using the storage.

I'm basically suggesting to add the following logic right after `if ( ! $this->is_user_admin() ) {` and once we have `$current_wp_user`:
```php
    $user_id = $current_wp_user->ID;

    if ( get_transient( "locked_{$user_id}" ) ){
         return;
    }

    set_transient( "locked_{$user_id}", true,  20); // 20-sec lock.
```

That way we make sure that each admin will only execute this method once in 20 sec, max.

includes/class-freemius.php
```diff
@@ -629,6 +636,254 @@ private function _version_updates_handler() {
...
+            $storage         = FS_Storage::instance( 'gdpr_global', '' );
+            $current_wp_user = self::_get_current_wp_user();
+
+            self::$_all_admin_notices = FS_Admin_Notices::instance( 'all_admins', '', '', true );
+
+            if ( self::$_all_admin_notices->has_sticky( "gdpr_optin_actions_{$current_wp_user->ID}", true ) ) {
+                return;
+            }
+
+            $show_gdpr_optin_notice = $storage->get( "show_gdpr_optin_notice_{$current_wp_user->ID}", true );
```
vovafeldman: @fajardoleo this is a gdpr focused storage, so there's no need to specify that in the var names. You can use `show_notice_{$current_wp_user->ID}` instead.

includes/class-freemius.php
```diff
@@ -629,6 +636,254 @@ private function _version_updates_handler() {
...
+            if ( ! $show_gdpr_optin_notice ) {
+                return;
+            }
+
+            $last_time_gdpr_optin_notice_shown = $storage->get( "gdpr_optin_notice_shown_{$current_wp_user->ID}", false );
```
vovafeldman: @fajardoleo same here, I'd suggest using `notice_shown_at_{$current_wp_user->ID}` instead.

includes/class-freemius.php
```diff
@@ -19058,6 +19316,204 @@ private function get_complete_upgrade_instructions( $plan_title = '' ) {
...
+                'body'    => $params,
+                'timeout' => WP_FS__DEBUG_SDK ? 60 : 30,
+            );
+
+            $result = array();
+
+            $total_plugin_ids             = count( $plugin_ids );
+            $plugin_ids_count_per_request = 10;
+            for ( $i = 1; $i <= $total_plugin_ids; $i += $plugin_ids_count_per_request ) {
+                $plugin_ids_set                = array_slice( $plugin_ids, $i - 1, $plugin_ids_count_per_request );
```
vovafeldman: @fajardoleo since this method isn't using our API layer, we need to add some caching to avoid a situation where for some reason this request is called multiple times in a short period. I think that transients are the easiest here. So you can do something like that:
```php
    $cache_key = 'fs_user_plugins_' . md5( $user_email, implode( ',', $plugin_ids_set) );
    $response = get_transient( $cache_key );

    if ( false === $response ) {
        ... actual API request ...
        set_transient( $cache_key, $response, WP_FS__TIME_24_HOURS_IN_SEC);
    }

    ... a logic that processes the response...
```

PR #256, opened 2018-09-03
[opt-in] [ms] Fixed the redirection after opting in or skipping the o…

fajardoleo:
…pt-in when the plugin doesn't have a network-level menu.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -14572,7 +14572,7 @@
 
             $hook = false;
 
-            if ( ! $this->_menu->has_menu() ) {
+            if ( ! $this->has_settings_menu() ) {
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -15003,6 +14994,14 @@ private function embed_submenu_items() {
                             array( $this, '' )
                         );
                     }
+
+                    if ( $item['show_submenu'] && ! $removed_first_submenu_item ) {
```
vovafeldman: @fajardoleo please add that Please add the explanation that we just discussed on slack including a “visual” example of what happens:
```
    Awesome Plugin
        - Awesome Plugin  <-- we want to remove this since there's no real setting page for the top-level
```

PR #257, opened 2018-09-06
[deactivation] [subscription] Enriching the deactivation feedback for…

fajardoleo:
…m to optionally support subscription cancellation options.


templates/forms/deactivation/form.php
```diff
@@ -399,22 +568,59 @@ function showPanel(panelType) {
...
+            $primaryButton.text('<?php echo esc_js( sprintf(
+                fs_text_inline( 'Proceed', 'proceed' , $slug )
+            ) ) ?>');
+
+            $secondaryButton.text('<?php echo esc_js( sprintf(
+                fs_text_inline( 'Cancel', 'cancel' , $slug )
+            ) ) ?>');
 		} else {
-			$deactivateButton.html('<?php echo esc_js( sprintf(
+			$secondaryButton.html('<?php echo esc_js( sprintf(
```
vovafeldman: @fajardoleo use `json_encode()`

assets/scss/admin/_deactivation-feedback.scss
```diff
@@ -48,6 +48,50 @@
             line-height: 1.5em;
         }
     }
+
+	&.has-subscription-actions {
```
vovafeldman: @fajardoleo this is relevant for all colors of this class, just like it's not a good practice to use random values in code, since we have vars in SASS it's better to use meaningful var names instead. So please add the colors to `_colors.scss` as variables under the `// WP Buttons.` section. Use the `$wp-button-alert-` prefix for the var names.

includes/class-freemius.php
```diff
@@ -2058,6 +2063,48 @@ function _submit_uninstall_reason_action() {
             exit;
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since  2.1.4
+         */
+        function cancel_subscription_or_trial_ajax_action() {
```
vovafeldman: @fajardoleo we need to add some error handling to the cancellation to avoid cases when a user cancels a subscription and later get billed because of some error.

includes/class-freemius.php
```diff
@@ -2058,6 +2063,48 @@ function _submit_uninstall_reason_action() {
...
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since  2.1.4
+         *
+         * @param number   $plugin_id
+         * @param int|null $blog_id
+         */
+        private function cancel_subscription_or_trial( $plugin_id, $blog_id = null ) {
```
vovafeldman: @fajardoleo I understand that you've used the `$blog_id` since you've leveraged this method to replace the pieces of logic that cancel the subscription in other places in the code, but while it may make sense there, as a standalone method it doesn't make any sense. Why would a method that cancels a subscription/trial will switch the DB cursor to the network's install main blog ID? and only if the given `$blog_id` is numeric. It just doesn't make any sense (unless I'm missing something). Therefore, remove this part from the method and just add it in the place(s) where it was triggering as an additional code that is executed after the method call.

includes/class-freemius.php
```diff
@@ -2058,6 +2063,48 @@ function _submit_uninstall_reason_action() {
...
+            if ( $plugin_id == $this->get_id() ) {
+                $fs = $this;
+            } else if ( $this->is_addon_activated( $plugin_id ) ) {
+                $fs = self::get_instance_by_id( $plugin_id );
+
+                $blog_id = null;
+            }
+
+            if ( ! is_null( $fs ) ) {
+                if ( $fs->is_trial() ) {
```
vovafeldman: @fajardoleo we only want to cancel the trial when it's a paid trial, i.e. check if `is_paid_trial()` (you may also need to adjust the logic of the dialog, we don't want to show the "cancel subscription" section if it's a trial without a payment method).

PR #262, opened 2018-09-30
[readme] [api] Retrieve the readme info for premium code type from th…

fajardoleo:
…e server.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -17425,7 +17425,13 @@
 
             // If add-on and not yet activated, try to fetch based on server licensing.
             if ( is_bool( $is_premium ) ) {
-                $endpoint = add_query_arg( 'is_premium', json_encode( $is_premium ), $endpoint );
+                $endpoint = add_query_arg(
+                    array(
+                        'is_premium' => json_encode( $is_premium ),
+                        'readme'     => 'true'
+                    ),
+                    $endpoint
+                );
             }
 
             if ( $this->has_secret_key() ) {
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -4076,6 +4076,8 @@
                 }
             }
 
+            $parent_plugin_id = fs_request_get( 'parent_plugin_id', false );
+
             /**
              * This should be executed even if Freemius is off for the core module,
              * otherwise, the add-ons dialogbox won't work properly. This is esepcially
@@ -4084,14 +4086,17 @@
              * @author Vova Feldman (@svovaf)
              */
             if ( $this->is_user_in_admin() &&
-                 ! $this->is_addon() &&
-                 $this->has_addons() &&
                  'plugin-information' === fs_request_get( 'tab', false ) &&
-                 $this->get_id() == fs_request_get( 'parent_plugin_id', false )
+                 false != $parent_plugin_id
             ) {
-                require_once WP_FS__DIR_INCLUDES . '/fs-plugin-info-dialog.php';
+                if (
+                    ( ! $this->is_addon() && $this->has_addons() && $this->get_id() == $parent_plugin_id ) ||
+                    ( $this->is_addon() && $this->get_parent_id() == $parent_plugin_id )
+                ) {
+                    require_once WP_FS__DIR_INCLUDES . '/fs-plugin-info-dialog.php';
 
-                new FS_Plugin_Info_Dialog( $this );
+                    new FS_Plugin_Info_Dialog( $this->is_addon() ? $this->get_parent_instance() : $this );
+                }
             }
 
             // Check if Freemius is on for the current plugin.
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -16323,11 +16323,12 @@
          * @param bool|number $plugin_id
          * @param bool        $flush      Since 1.1.7.3
          * @param int         $expiration Since 1.2.2.7
+         * @param bool|string $newer_than Since 2.2.1
          *
          * @return object|false New plugin tag info if exist.
          */
-        private function _fetch_newer_version( $plugin_id = false, $flush = true, $expiration = WP_FS__TIME_24_HOURS_IN_SEC ) {
-            $latest_tag = $this->_fetch_latest_version( $plugin_id, $flush, $expiration );
+        private function _fetch_newer_version( $plugin_id = false, $flush = true, $expiration = WP_FS__TIME_24_HOURS_IN_SEC, $newer_than = false ) {
+            $latest_tag = $this->_fetch_latest_version( $plugin_id, $flush, $expiration, $newer_than );
 
             if ( ! is_object( $latest_tag ) ) {
                 return false;
@@ -16352,17 +16353,18 @@
          * @param bool|number $plugin_id
          * @param bool        $flush      Since 1.1.7.3
          * @param int         $expiration Since 1.2.2.7
+         * @param bool|string $newer_than Since 2.2.1
          *
          * @return bool|FS_Plugin_Tag
          */
-        function get_update( $plugin_id = false, $flush = true, $expiration = WP_FS__TIME_24_HOURS_IN_SEC ) {
+        function get_update( $plugin_id = false, $flush = true, $expiration = WP_FS__TIME_24_HOURS_IN_SEC, $newer_than = false ) {
             $this->_logger->entrance();
 
             if ( ! is_numeric( $plugin_id ) ) {
                 $plugin_id = $this->_plugin->id;
             }
 
-            $this->check_updates( true, $plugin_id, $flush, $expiration );
+            $this->check_updates( true, $plugin_id, $flush, $expiration, $newer_than );
             $updates = $this->get_all_updates();
 
             return isset( $updates[ $plugin_id ] ) && is_object( $updates[ $plugin_id ] ) ? $updates[ $plugin_id ] : false;
@@ -17472,13 +17474,7 @@
 
             // If add-on and not yet activated, try to fetch based on server licensing.
             if ( is_bool( $is_premium ) ) {
-                $endpoint = add_query_arg(
-                    array(
-                        'is_premium' => json_encode( $is_premium ),
-                        'readme'     => json_encode( $is_premium )
-                    ),
-                    $endpoint
-                );
+                $endpoint = add_query_arg( 'is_premium', json_encode( $is_premium ), $endpoint );
             }
 
             if ( $this->has_secret_key() ) {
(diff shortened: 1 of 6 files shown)


includes/class-fs-plugin-updater.php
```diff
@@ -183,15 +245,41 @@ function edit_and_echo_plugin_update_row( $file, $plugin_data ) {
...
-                '/(\<div.+>)(.+)(\<a.+\<a.+)\<\/div\>/is',
-                '$1 $2 ' . sprintf(
-                    $this->_fs->get_text_inline( '%sRenew your license now%s to access version %s security & feature updates, and support.', 'renew-license-now' ),
-                    '<a href="' . $this->_fs->pricing_url() . '">', '</a>',
-                    $r->new_version ) .
-                '$4',
-                $plugin_update_row
-            );
+            if ( ! $this->_fs->has_any_active_valid_license() ) {
+                $plugin_update_row = preg_replace(
```
vovafeldman: @fajardoleo same here, please add an explanation of what this regex does. If possible, include an example of an input and output.

includes/class-fs-plugin-updater.php
```diff
@@ -401,6 +486,18 @@ function get_update_details( FS_Plugin_Tag $new_version ) {
                 );
             }
 
+            if ( $this->_fs->is_premium() ) {
+                $latest_tag = $this->_fs->_fetch_latest_version( $this->_fs->get_id(), false );
```
vovafeldman: @fajardoleo I think that I'm missing something... based on the API PR, the endpoint was enriched with a new `readme` param that defaults to `false`. It needs to be `true` to return the readme data. I don't see any changes to `_fetch_latest_version()` so I don't understand how it will ever return the data with the readme.

includes/class-fs-plugin-updater.php
```diff
@@ -769,6 +866,52 @@ function plugins_api_filter( $data, $action = '', $args = null ) {
 
                 $data->version       = $new_version->version;
                 $data->download_link = $new_version->url;
+
+                if ( isset( $new_version->readme ) && is_object( $new_version->readme ) ) {
```
vovafeldman: @fajardoleo same comment as the previous, I don't see any changes to `get_latest_download_details()` so I don't understand how it will ever return the data with the readme.

includes/fs-plugin-info-dialog.php
```diff
@@ -64,7 +64,7 @@ function __construct( Freemius $fs ) {
         function _get_addon_info_filter( $data, $action = '', $args = null ) {
             $this->_logger->entrance();
 
-            $parent_plugin_id = fs_request_get( 'parent_plugin_id', false );
+            $parent_plugin_id = fs_request_get( 'parent_plugin_id', $this->_fs->get_id() );
```
vovafeldman: @fajardoleo can you please explain this change?

templates/forms/premium-versions-upgrade-handler.php
```diff
@@ -20,11 +20,30 @@
...
+                WP_FS__PERIOD_LIFETIME,
+            false,
+            array( 'licenses' => $license->quota )
+        );
+    }
+
     $message = sprintf(
         fs_text_inline( 'There is a new version of %s available.', 'new-version-available-message', $slug ) .
-        fs_text_inline( ' %sRenew your license now%s to access version %s security & feature updates, and support.', 'renew-license-now', $slug ),
+        fs_text_inline( ' %s%s now%s to access version %s security & feature updates, and support.', 'renew-license-now', $slug ),
```
vovafeldman: @fajardoleo same comment as before regarding the translatable string (just include the `now` as part of the strings). Also, please notice that you've used `'buy-a-license'` for the key of the `''Buy a license'` string where previously it was `'buy-license'`. We need to be consistent across the SDK to use the same keys when the strings are identical.

includes/class-fs-plugin-updater.php
fajardoleo: @vovafeldman, `get_latest_download_details()` calls `_fetch_latest_version()` which in turn calls `_get_latest_version_endpoint()` which is responsible for the inclusion of the `readme` param.

includes/class-fs-plugin-updater.php
fajardoleo: @vovafeldman, `_fetch_latest_version()` calls `_get_latest_version_endpoint()` which is responsible for the inclusion of the `readme` param.

includes/class-freemius.php
```diff
@@ -11365,6 +11378,40 @@ function get_trial_url() {
...
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.1.4
+         *
+         * @param string $new_version
+         *
+         * @return string
+         */
+        function version_upgrade_checkout_link( $new_version ) {
```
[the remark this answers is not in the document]
fajardoleo: @vovafeldman, when it's private, it's not working. May I know if you meant something else?

includes/fs-plugin-info-dialog.php
fajardoleo: @vovafeldman, previously, the custom plugin information dialog for add-ons was showing up only when triggered from the "Add-Ons" page. This change is for also showing the custom plugin information dialog for add-ons from the "Plugins" and "WP Admin -> Dashboard -> Updates" pages.

includes/class-freemius.php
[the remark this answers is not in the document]
vovafeldman: @fajardoleo I didn't notice it is used in `class-fs-plugin-updater.php`.

templates/forms/premium-versions-upgrade-handler.php
```diff
@@ -20,20 +20,42 @@
...
+                WP_FS__PERIOD_LIFETIME,
+            false,
+            array( 'licenses' => $license->quota )
+        );
+    }
+
     $message = sprintf(
         fs_text_inline( 'There is a new version of %s available.', 'new-version-available-message', $slug ) .
-        fs_text_inline( ' %sRenew your license now%s to access version %s security & feature updates, and support.', 'renew-license-now', $slug ),
+        fs_text_inline( ' %s to access version %s security & feature updates, and support.', 'renew-license-now', $slug ),
```
vovafeldman: @fajardoleo `'renew-license-now'` is now conflicting since it's also used as the string key here:
```
fs_text_inline( 'Renew your license now', 'renew-license-now', $slug )
```

I suggest renaming the key to `'x-for-updates-and-support'`.

includes/class-freemius.php
```diff
@@ -11365,6 +11378,40 @@ function get_trial_url() {
...
+                        WP_FS__PERIOD_LIFETIME,
+                    false,
+                    array( 'licenses' => $this->_license->quota )
+                );
+
+                $purchase_license_text = $this->get_text_inline( 'Renew your license', 'renew-your-license' );
+            }
+
+            return sprintf(
+                $this->get_text_inline( '%s to access version %s security & feature updates, and support.', 'renew-license-now' ),
```
vovafeldman: @fajardoleo `'renew-license-now'` is now conflicting since it's also used as the string key here:
```
fs_text_inline( 'Renew your license now', 'renew-license-now', $slug )
```

I suggest renaming the key to `'x-for-updates-and-support'`.

includes/class-freemius.php
```diff
@@ -11365,6 +11378,40 @@ function get_trial_url() {
...
+
+                $url = $this->checkout_url(
+                    is_object( $subscription ) ?
+                        ( 1 == $subscription->billing_cycle ? WP_FS__PERIOD_MONTHLY : WP_FS__PERIOD_ANNUALLY ) :
+                        WP_FS__PERIOD_LIFETIME,
+                    false,
+                    array( 'licenses' => $this->_license->quota )
+                );
+
+                $purchase_license_text = $this->get_text_inline( 'Renew your license', 'renew-your-license' );
```
vovafeldman: @fajardoleo for consistency, I suggest to use the same string here:
```
fs_text_inline( 'Renew your license now', 'renew-license-now', $slug )
```

includes/class-fs-plugin-updater.php
```diff
@@ -120,6 +120,75 @@ private function filters() {
...
+            $license = $this->_fs->_get_license();
+
+            $subscription = ( is_object( $license ) && ! $license->is_lifetime() ) ?
+                $this->_fs->_get_subscription( $license->id ) :
+                null;
+
+            $contents = ob_get_clean();
+
+            /**
+             * Replace the plugin information dialog's "Install Update Now" button's text and URL. If there's a license,
```
vovafeldman: @fajardoleo thanks for the explanation. Question -> why do we "hack" it instead of modifying it in the `fs-plugin-info-dialog.php`? As far as I know, we have a full control over the rendering of the plugin's info dialog.

includes/class-freemius.php
```diff
@@ -17425,7 +17472,13 @@ private function _get_latest_version_endpoint( $addon_id = false, $type = 'json'
 
             // If add-on and not yet activated, try to fetch based on server licensing.
             if ( is_bool( $is_premium ) ) {
-                $endpoint = add_query_arg( 'is_premium', json_encode( $is_premium ), $endpoint );
+                $endpoint = add_query_arg(
+                    array(
+                        'is_premium' => json_encode( $is_premium ),
+                        'readme'     => json_encode( $is_premium )
```
vovafeldman: @fajardoleo did you verify that we actually need the readme on every request when running from a premium context? The readme adds quite a lot of content to the result (and bandwidth to our servers).

For example, when `_fetch_newer_version()` executed and there's no newer version, we don't really need to return the readme. Maybe to avoid duplicate calls (once to check if there's a newer version and then another to get the readme), we can enrich the endpoint with a `newer_than` param which will receive the current's product version and will only return a result when the latest matching version actually newer than the one specified by `newer_than`? (behave like that only if that param is specified

includes/class-fs-plugin-updater.php
fajardoleo: @vovafeldman, `fs-plugin-info-dialog.php` is used only for add-ons.

includes/class-freemius.php
```diff
@@ -17495,13 +17491,17 @@ private function _get_latest_version_endpoint( $addon_id = false, $type = 'json'
...
          *
          * @return object|false Plugin latest tag info.
          */
         function _fetch_latest_version(
             $addon_id = false,
             $flush = true,
-            $expiration = WP_FS__TIME_24_HOURS_IN_SEC
+            $expiration = WP_FS__TIME_24_HOURS_IN_SEC,
+            $newer_than = false,
+            $readme = true
```
vovafeldman: @fajardoleo I'd recommend to refactor it to `$include_readme` or `$fetch_readme`.

includes/class-fs-plugin-updater.php
```diff
@@ -941,11 +944,13 @@ function plugins_api_filter( $data, $action = '', $args = null ) {
...
          *
          * @param number|bool $addon_id
+         * @param bool|string $newer_than Since 2.2.1
+         * @param bool|string $readme     Since 2.2.1
          *
          * @return object
          */
-        private function get_latest_download_details( $addon_id = false ) {
-            return $this->_fs->_fetch_latest_version( $addon_id );
+        private function get_latest_download_details( $addon_id = false, $newer_than = false, $readme = true ) {
```
vovafeldman: @fajardoleo I'd recommend to refactor it to `$include_readme` or `$fetch_readme`.

includes/class-freemius.php
fajardoleo: Refactored to `$fetch_readme`.

includes/class-fs-plugin-updater.php
fajardoleo: Refactored to `$fetch_readme`.

Issue #266, opened 2018-10-04
Child Theme Support broken upon change of state from free to premium (-premium suffix)


Luehrsen:
**Actual Behavior**:
When designing a child theme upon a freemius theme it is required to define the theme folder slug for the parent theme in the stylesheet header of the child theme. (see https://codex.wordpress.org/Child_Themes)

It is expected behaviour in freemius that the folder name changes, but that makes it impossible to create a child theme that works in any state of the freemius theme. (free and premium)

**Versions**: (*)
 -  `Freemius SDK Version:` 2.1.2
 -  `WordPress Version:` 4.9
 -  `PHP Version:` 7.1

**Theme**: (*)
 -  `Name:` Agncy
 -  `Slug:` agncy
 -  `Freemius ID:` 1282

**Additional Information**:
 -  `Browser Type: (*)` Chrome
 -  `Browser Version: (*)` 69.0.3497.100
 -  `OS: (*)` MacOS 10.14

Luehrsen:
This is the check if the parent theme can be found, based upon the stylesheet header information.

https://github.com/WordPress/WordPress/blob/56c162fbc9867f923862f64f1b4570d885f1ff03/wp-includes/class-wp-theme.php#L342

Luehrsen:
This is where the parent theme gets loaded in a child theme context.

https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-theme.php#L370

vovafeldman:
@fajardoleo please take a look at this one. I think that the easiest solution would be to tap to WP logic and make it search for the `-premium` folder when running the premium theme version. I'm not sure if that's possible, but if it is, that's probably the easiest for now.

fajardoleo:
Hi @Luehrsen, please check this out and let us know if it resolves the issue:
https://github.com/Freemius/wordpress-sdk/pull/268

If it doesn't work at first, please switch to a different theme then switch back to the child theme.

vovafeldman:
@Luehrsen do you have control of the child theme? Can we propose a fix by adding code to the child theme?

Otherwise, due to the way child themes are loaded, if there are no plugins that running the Freemius SDK, the only option that will make it work is if the user will first activate the parent theme (to manipulate the core mechanism that identifies the relevant parent template), and only then switch to the child. Also, if the child theme will be later switched to a different theme (without Freemius), after 30 min (that's when the cache expires), WP will show an error that the child's parent theme is missing.

[...]

PR #267, opened 2018-10-10
[theme] Handling of the case when a child theme's parent is a free th…

fajardoleo:
…eme version but the installed theme is the premium one.


vovafeldman:
Newer PR #266

PR #268, opened 2018-10-11
[theme] Handling of the case when a child theme's parent is a free th…

fajardoleo:
…eme version but the installed theme is the premium one (cache manipulation).


Luehrsen:
Alright. This code snippet works* and we are currently shipping it with our theme. 

Works is relative here, because if you do not have the parent theme active (but the -premium version installed), the child theme will be shown to have no parent theme. This might throw off some users, but that is certainly catchable in support.

PR #271, opened 2018-10-16
[license] [activation] After an activation of a premium module, force…

fajardoleo:
… the user to enter a license, even with freemium products.


vovafeldman:
@fajardoleo after reviewing the code, I understand that your implementation will basically require the license activation on once. If the user deactivates a license, it will not require another license activation. I didn't think about it, but this is actually a better UX, maybe except when the license is blocking (but let's leave this use-case aside for now). Long story short, I think that it will make the code clearer if instead of `skip_license_activation` we actually use the opposite `require_license_activation`.

PR #278, opened 2018-11-01
[cache] Clear the plugin data cache if the trigger to add it was the …

fajardoleo:
…FS SDK in order to not affect custom headers that may be added by other plugins.

### DIFF
--- a/includes/fs-essential-functions.php
+++ b/includes/fs-essential-functions.php
@@ -61,7 +61,7 @@
 
                 $plugins = get_plugins();
 
-                if ( $delete_cache && isset( $plugins['woocommerce/woocommerce.php'] ) ) {
+                if ( $delete_cache && is_plugin_active( 'woocommerce/woocommerce.php' ) ) {
                     wp_cache_delete( 'plugins', 'plugins' );
                 }
             }
--- a/includes/fs-essential-functions.php
+++ b/includes/fs-essential-functions.php
@@ -36,6 +36,39 @@
 		}
 	}
 
+	if ( ! function_exists( 'fs_get_plugins' ) ) {
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.2.1
+         *
+         * @param string $plugin_folder
+         * @param bool   $delete_cache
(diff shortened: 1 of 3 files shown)


includes/fs-essential-functions.php
```diff
@@ -36,6 +36,39 @@ function fs_normalize_path( $path ) {
...
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.2.1
+         *
+         * @param string $plugin_folder
+         * @param bool   $delete_cache
+         *
+         * @return array
+         */
+	    function fs_get_plugins( $plugin_folder = '', $delete_cache = false ) {
```
vovafeldman: @fajardoleo since all the calls to `fs_get_plugins()` always use an empty string for the 1st param, there's no need to have it at all. If we'll need to use it in the future, we'll add it.

includes/fs-essential-functions.php
```diff
@@ -36,6 +36,39 @@ function fs_normalize_path( $path ) {
...
+            if ( ! $cached_plugins = wp_cache_get( 'plugins', 'plugins' ) ) {
```
vovafeldman: @fajardoleo this could be quite confusing together with the `!`. I would prefer if you move the assignment outside the `if`.

includes/fs-essential-functions.php
```diff
@@ -36,6 +36,39 @@ function fs_normalize_path( $path ) {
...
+            }
+
+            if ( isset( $cached_plugins[ $plugin_folder ] ) ) {
+                $plugins = $cached_plugins[ $plugin_folder ];
+            } else {
+                if ( ! function_exists( 'get_plugins' ) ) {
+                    require_once ABSPATH . 'wp-admin/includes/plugin.php';
+                }
+
+                $plugins = get_plugins();
```
vovafeldman: @fajardoleo from what I've seen, `get_plugins()` is a resource intensive method, therefore, I suggest that before we delete the cache, first check if WooCommerce is installed at all. And only clear it if it is installed.

PR #279, opened 2018-11-01
Added `premium_slug` and `premium_suffix` integration settings that a…

fajardoleo:
…re used in customizing the premium version's folder name and title suffix.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -8277,6 +8277,8 @@
         }
 
         /**
+         * Set the basename of the current product and hook _activate_plugin_event_hook() to the activation action.
+         *
          * @author Vova Feldman (@svovaf)
          * @since  2.2.1
          *
@@ -8477,6 +8479,8 @@
         }
 
         /**
+         * Retrieve the desired folder name for the product.
+         *
          * @author Vova Feldman (@svovaf)
          * @since  1.2.1.7
          *
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -8314,19 +8314,19 @@
         /**
          * @author Vova Feldman (@svovaf)
          * @since  1.1.1
+         * @since  2.2.1 If the context product is in its premium version, use the current module's basename, even if it was renamed.
          *
          * @return string
          */
         function premium_plugin_basename() {
-            if ( isset( $this->_premium_plugin_basename ) ) {
-                return $this->_premium_plugin_basename;
+            if ( ! isset( $this->_premium_plugin_basename ) ) {
+                $this->_premium_plugin_basename = $this->is_premium() ?
+                    // The product is premium, so use the current basename.
+                    $this->_plugin_basename :
+                    $this->get_premium_slug() . '/' . basename( $this->_free_plugin_basename );
             }
 
-            if ( $this->is_premium() ) {
-                return $this->_plugin_basename;
-            }
-
-            return "{$this->get_premium_slug()}/" . basename( $this->_free_plugin_basename );
+            return $this->_premium_plugin_basename;
         }
 
         /**
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -381,7 +381,7 @@
             $this->_plugin_main_file_path = $this->_find_caller_plugin_file( $is_init );
             $this->_plugin_dir_path       = plugin_dir_path( $this->_plugin_main_file_path );
             $this->_plugin_basename       = $this->get_plugin_basename();
-            $this->_free_plugin_basename  = ( $this->_slug . '/' . $this->_plugin_basename );
+            $this->_free_plugin_basename  = ( $this->_slug . '/' . basename( $this->_plugin_basename ) );
 
             $this->_is_multisite_integrated = (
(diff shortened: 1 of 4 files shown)


includes/class-freemius.php
```diff
@@ -4739,24 +4747,42 @@ private function parse_settings( &$plugin_info ) {
...
+                'title'                => $this->get_plugin_name( $premium_suffix ),
                 'file'                 => $this->_plugin_basename,
                 'is_premium'           => $this->get_bool_option( $plugin_info, 'is_premium', true ),
+                'premium_suffix'       => $premium_suffix,
                 'is_live'              => $this->get_bool_option( $plugin_info, 'is_live', true ),
                 'affiliate_moderation' => $this->get_option( $plugin_info, 'has_affiliation' ),
             ) );
 
             if ( $plugin->is_updated() ) {
+                $register_activation_hook = ( ! is_object( $this->_plugin ) || $this->_plugin->premium_slug !== $premium_slug );
```
vovafeldman: @fajardoleo: 
1. I don't understand this addition, can you please explain it?
2. Also, the name of the method is `parse_settings()` so the registration of the activation hook should not be here anyway.

includes/class-freemius.php
```diff
@@ -381,7 +381,7 @@ private function __construct( $module_id, $slug = false, $is_init = false ) {
             $this->_plugin_main_file_path = $this->_find_caller_plugin_file( $is_init );
             $this->_plugin_dir_path       = plugin_dir_path( $this->_plugin_main_file_path );
             $this->_plugin_basename       = $this->get_plugin_basename();
-            $this->_free_plugin_basename  = str_replace( '-premium/', '/', $this->_plugin_basename );
+            $this->_free_plugin_basename  = ( $this->_slug . '/' . basename( $this->_plugin_basename ) );
```
vovafeldman: @fajardoleo I think that this change is a little risky and things may work unexpectedly if the admin changes the free version's folder name. We may want to update it in `parse_settings()` after loading the info if we are running the free or paid product, and doing something like that:
```
if ( ! $this->is_premium() ) {
    $this->_free_plugin_basename = $this->_plugin_basename;
}
```

But we need to verify that the `_free_plugin_basename` isn't actually used in the logic between the constructor and the parsing of the settings.

includes/class-freemius.php
```diff
@@ -8272,7 +8298,7 @@ function _uninstall_plugin_event( $check_user = true ) {
          * @return string
          */
         function premium_plugin_basename() {
-            return "{$this->_slug}-premium/" . basename( $this->_free_plugin_basename );
+            return "{$this->get_premium_slug()}/" . basename( $this->_free_plugin_basename );
```
vovafeldman: @fajardoleo worth checking if we can change this to:
```
    return ( $this->is_premium() ) ?
        $this->_plugin_basename :
        "{$this->get_premium_slug()}/" . basename( $this->_free_plugin_basename );
```

I hope that this will protect against situations where renaming the folder of the premium version. I think that we'll need to enrich `is_premium()` for cases where `$this->_plugin->is_premium` is not set.

includes/class-freemius.php
fajardoleo: @vovafeldman, when the settings are parsed and the premium slug doesn't match the stored one, the activation hook needs to be registered again since the previous registration logic doesn't have the correct premium version's directory name anymore (`_register_hooks` is called directly in the constructor and it registers an activation hook before the settings are parsed).

includes/class-freemius.php
```diff
@@ -2186,6 +2192,16 @@ static function instance( $module_id, $slug = false, $is_init = false ) {
             return self::$_instances[ $key ];
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.2.1
+         *
+         * @return Freemius[]
+         */
+        static function instances() {
```
vovafeldman: See if we can get rid of this.

includes/class-freemius.php
```diff
@@ -8488,19 +8566,27 @@ function has_secret_key() {
...
          * @return string
          */
-        function get_plugin_name() {
+        function get_plugin_name( $premium_suffix = false ) {
             $this->_logger->entrance();
 
-            if ( ! isset( $this->_plugin_name ) ) {
+            if (
+                ! isset( $this->_plugin_name ) ||
+                (
```
vovafeldman: I don't think there's a point for this part of the `if` since once `$this->_plugin_name` is set, we never unset it.

includes/class-fs-plugin-updater.php
```diff
@@ -1078,13 +1108,56 @@ private function is_premium_plugin_active( $plugin_id = false ) {
          *
          * @return string|WP_Error
          */
-        function _maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
+        static function _maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
```
vovafeldman: I don't understand the reasoning converting this method to static. On the contrary, it looks like it make things more complex since you don't have a direct access to the `$this->_fs` object. Am I missing something?

PR #289, opened 2018-12-05
[bug-fix] Php warning after plugin install

bradvin:
There was a PHP warning after installing a plugin or theme by uploading a .zip


vovafeldman:
Thanks for the patch, @bradvin!

PR #308, opened 2019-03-01
Overly Cautious Sanitization to Stop False Flags

Ipstenu:
Whenever we run a scan of a plugin using Freemius, we have to double check that this is safe.

The answer is: It could be SAFER and stop throwing red-flags if you just act extra paranoid.

It would benefit users as well to see that you're going that extra mile to make ABSOLUTELY sure people can't do stupid things to the code library they're using.

Yes, it's overkill, but I firmly believe that staying one half-step ahead of hackers means being paranoid.

### DIFF
--- a/includes/fs-core-functions.php
+++ b/includes/fs-core-functions.php
@@ -142,13 +142,13 @@
 
             switch ( $type ) {
                 case 'post':
-                    $value = isset( $_POST[ $key ] ) ? $_POST[ $key ] : $def;
+                    $value = isset( $_POST[ $key ] ) ? sanitize_text_field( $_POST[ $key ] ) : $def;
                     break;
                 case 'get':
-                    $value = isset( $_GET[ $key ] ) ? $_GET[ $key ] : $def;
+                    $value = isset( $_GET[ $key ] ) ? sanitize_text_field( $_GET[ $key ] ) : $def;
                     break;
                 default:
-                    $value = isset( $_REQUEST[ $key ] ) ? $_REQUEST[ $key ] : $def;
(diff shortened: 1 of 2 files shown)


vovafeldman:
Thanks for the PR, @Ipstenu. These are helper functions we use for accessing GET/POST params with an optional default value. The actual sanitization is done in the scope of the function's usage.

`sanitize_text_field()` triggers the filters mechanism and the sanitzation logic executes several `regex` related calls which are known to be "hungry" on resources. So it will definitely have an impact on performance.

Maybe there's some other way that we can flag/whitelist that code for reviewers? Or maybe there's a similar core function that we can use instead?

If not, we would first need to review all the logic that calls those functions and probably introduce several new helper functions for various use-cases. I already know several places where the exepcted value is an array of strings (and not a string), so accepting the PR as is will break the logic because `wp_check_invalid_utf8()` expects a string and for example, will throw an `E_NOTICE` when trying to cast an array to a string:
```
function wp_check_invalid_utf8( $string, $strip = false ) {
	$string = (string) $string;

	if ( 0 === strlen( $string ) ) {
		return '';
	}

        ...
```

Ipstenu:
The rule of thumb is you sanitize EARLY. You’re doing it very late, which is why it gets caught every time. At best, if you could use something like phpcs comments around the troublesome code to indicate what it is and why it’s okay that would help. Then I wouldn’t have to backtrack and make SURE this is the secure code and not something someone’s daftly edited.

But it’s still terrible practice to be passing unsanitized code around if you don’t have to. And frankly we can’t assume that your library is untouched by developers. You know they do stupid things. They also do malicious things, and insert hacks into ‘safe’ libraries, so it’s not acceptable for us to ignore your library and trust it’s safe. Not because of YOU but because of people who use it 😕

PR #317, opened 2019-03-24
Feature/premium theme details dialog

fajardoleo:


includes/fs-plugin-info-dialog.php
```diff
@@ -955,18 +982,27 @@ function install_plugin_information() {
                 }
             }
 
-            // Add after $api->slug is ready.
-            $plugins_section_titles['features'] = fs_text_x_inline( 'Features & Pricing', 'Plugin installer section title', 'features-and-pricing', $api->slug );
+            if ( ! $is_theme ) {
+                // Add after $api->slug is ready.
+                $plugins_section_titles['features'] = fs_text_x_inline( 'Features & Pricing', 'Plugin installer section title', 'features-and-pricing', $api->slug );
```
vovafeldman: @fajardoleo any reason we include the features & pricing tab only for plugins and not for themes?

includes/fs-plugin-info-dialog.php
```diff
@@ -1450,7 +1507,7 @@ class="fs-annual-discount"><?php printf(
             echo "</div>\n";
             echo "</div>\n";
             echo "</div>\n"; // #plugin-information-scrollable
-            echo "<div id='$tab-footer'>\n";
+            echo "<div id='$_tab-footer'>\n";
```
vovafeldman: @fajardoleo better to wrap the var with `{}`

fajardoleo: Updated

includes/fs-plugin-info-dialog.php
fajardoleo: Updated (the tab is relevant only to add-ons)

PR #319, opened 2019-03-28
[opt-in] [activation-mode] [fix]

fajardoleo:


includes/class-freemius.php
```diff
@@ -20220,7 +20223,7 @@ function _add_tracking_links() {
                 }
 
                 add_action( 'admin_footer', array( &$this, '_add_optout_dialog' ) );
-            } else {
+            } else if ( $this->is_anonymous() || $this->is_activation_mode() ) {
```
vovafeldman: @fajardoleo can you please explain this one?

PR #320, opened 2019-04-05
Feature/connected addon info enhancements

fajardoleo:


templates/account.php
```diff
@@ -590,14 +590,100 @@ class="fs-tag fs-<?php echo $fs->can_use_premium_code() ? 'success' : 'warn' ?>"
...
+
 											foreach ( $addons_to_show as $addon_id ) {
+                                                $addon              = $fs->get_addon( $addon_id );
+                                                $slug               = $addon->slug;
+											    $is_addon_connected = false;
+											    $addon_info         = array();
+
+                                                if ( isset( $sites[ $slug ] ) ) {
+                                                    $site   = $sites[ $slug ];
+                                                    $plugin = FS_Plugin_Manager::instance( $addon_id )->get();
```
vovafeldman: @fajardoleo are you sure that `FS_Plugin_Manager::instance( $addon_id )` will always return the `FS_Plugin` instance when this line is executed? If not, we need to address the case when the result is `false`, if yes, please explain why.

vovafeldman: @fajardoleo I feel that there's a confusion in the var names. For example, `$fs->get_addon( $addon_id )` also returns an `FS_Plugin` instance just like `FS_Plugin_Manager::instance( $addon_id )->get()`, so in the case of an active add-on, both values will be identical but the vars are inconsistent (`$addon` vs `$plugin). You also added `foreach`s that that use `$plan`, `$license`, and `$subscription` for the internal loop vars which will override the template's global values (not a good practice). I think that it worth trying to move all this new logic to a helper function.

PR #324, opened 2019-04-12
[add-on] Fixed issues related to installation of free add-on versions.

fajardoleo:


includes/class-freemius.php
```diff
@@ -9841,6 +9841,25 @@ function get_addon_by_slug( $slug, $flush = false ) {
...
+         *
+         * @return array
+         */
+        function _get_plans_and_pricing_by_addon_id() {
+            $result = $this->get_api_plugin_scope()->get( $this->add_show_pending( "/addons/pricing.json?type=visible" ) );
+
+            $plans_and_pricing_by_addon_id = array();
+            if ( $this->is_api_result_object( $result, 'addons' ) ) {
+                foreach ( $result->addons as $addon ) {
+                    $plans_and_pricing_by_addon_id[ $addon->id ] = $addon->plans;
```
vovafeldman: @fajardoleo please convert the plans to FS_Plugin_Plan entities.

includes/class-freemius.php
```diff
@@ -9841,6 +9841,25 @@ function get_addon_by_slug( $slug, $flush = false ) {
             return false;
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.2.5.1
+         *
+         * @return array
+         */
+        function _get_plans_and_pricing_by_addon_id() {
```
vovafeldman: @fajardoleo the name is confusing because it sounds like the method expects an addon ID. I think that it would be better to name it `_get_addons_plans_and_pricing_map_by_id()`

PR #334, opened 2019-05-02
Fixed options manager cache causes rewrite of options cache on every request.

sun:
Problem
* Freemius re-sets an option named "fs_cache_843" on every request with an empty array, which causes the total cache of all options to be autoloaded to be regenerated on every single page view.

Cause
* The existing code tries to manage a cache for performance already, but it considers an existing empty array value to need updating. The default value is empty though.

Proposed solution
1. Only update the stored options if there no existing ones yet.

Notes
* Bugfix was originally filed against: https://github.com/fooplugins/foogallery/pull/152

### DIFF
--- a/includes/managers/class-fs-option-manager.php
+++ b/includes/managers/class-fs-option-manager.php
@@ -153,7 +153,7 @@
 
                 $cached = true;
 
-                if ( empty( $this->_options ) ) {
+                if ( false === $this->_options ) {
                     if ( $this->_is_network_storage ) {
                         $this->_options = get_site_option( $option_name );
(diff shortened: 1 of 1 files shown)


vovafeldman:
Great catch, @sun! I reviewed the PR and it looks good. I'm happy to merge it though I'll need you to resubmit it against the `develop` branch. Or alternatively, we can just make the commit with your fix directly to `develop`. Up to you :) Just lmk

sun:
@vovafeldman Great! Rebased against `develop` 👍

vovafeldman:
Merged. Thanks for contributing, @sun!

PR #360, opened 2019-07-29
[trial] When determining if there's a paid trial context, also check …

fajardoleo:
…if there's an active subscription that is associated with the license.


vovafeldman:
Due to EOLs/spaces misalignment, the same changes were committed here: https://github.com/Freemius/wordpress-sdk/commit/c20575301513d85e89c12c5e852567c3fe043c4b

PR #364, opened 2019-08-27
Feature/bundle license activation enhancements

fajardoleo:


includes/class-freemius.php
```diff
@@ -7952,17 +8004,186 @@ private function maybe_network_activate_addon_license( $license = null ) {
...
+         * @param string|null $license_key
+         * @param bool        $flush
+         *
          * @return FS_Plugin_License
          */
-        private function get_addon_active_parent_license() {
+        function get_active_parent_license( $license_key = null, $flush = true ) {
             $parent_licenses_endpoint = "/plugins/{$this->get_id()}/parent_licenses.json?filter=activatable";
-            $parent_instance          = $this->get_parent_instance();
+            $parent_instance          = $this->is_addon() ?
```
vovafeldman, line 8178: @fajardoleo I would suggest to simplify the logic that calculates `$fs` as following:
```
$fs = $this;
if ( $this->is_addon() ) {
    $parent_instance = $this->get_parent_instance();

    if ( is_object( $parent_instance ) && $parent_instance->is_registered() ) {
        $fs = $parent_instance;
    }
}
```

includes/class-freemius.php
```diff
@@ -7986,9 +8215,24 @@ private function get_addon_active_parent_license() {
                 return null;
             }
 
-            $license = new FS_Plugin_License( $result->licenses[ 0 ] );
+            $parent_license = null;
 
-            return $license;
+            if ( empty( $license_key ) ) {
+                $parent_license = $result->licenses[ 0 ];
```
vovafeldman, line 8221: @fajardoleo is the space addition surrounding the `0` in the square brackets a WP convention?

includes/class-freemius.php
```diff
@@ -13076,15 +13346,16 @@ function activate_migrated_license(
...
             $plugin_id = null,
+            $sites = array(),
             $blog_id = null
         ) {
             $this->_logger->entrance();
 
             $result = $this->activate_license(
                 $license_key,
-                $this->is_network_active() ?
+                ( empty( $sites ) && is_null( $blog_id ) && $this->is_network_active() ) ?
```
vovafeldman, line 13356: @fajardoleo the previous logic was not validating that the `$blog_id` is null, meaning, it may not be backward compatible. Are you confident about this change?

includes/class-freemius.php
```diff
@@ -17167,10 +17444,11 @@ private function create_installs_with_user(
          * @author Vova Feldman (@svovaf)
          * @since  1.0.6
          *
-         * @param Freemius      $parent_fs
-         * @param bool|int|null $network_level_or_blog_id True for network level opt-in and integer for opt-in for specified blog in the network.
+         * @param Freemius          $parent_fs
+         * @param bool|int|null     $network_level_or_blog_id True for network level opt-in and integer for opt-in for specified blog in the network.
+         * @param FS_Plugin_License $bundle_license
```
vovafeldman, line 17449: @fajardoleo when we introduce new params to already existing methods it's a good practice to add `Since v<version>.` next to the parameter PHP Doc. Also, when the length of the params with the function is long, better to move each param to its own line.

package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Freemius, Inc.",
   "license": "GPL-3.0",
   "homepage": "https://freemius.com",
-  "version": "2.3.2",
+  "version": "2.3.1",
```
vovafeldman, line 7: @fajardoleo needs an update to `2.4.0`. Also, please update all the `@version` of new methods added in this PR to `2.4.0` as well (that will be the new release version since it's a major one).

templates/account.php
```diff
@@ -163,26 +163,55 @@
         }
     }
 
-    $is_child_license    = ( is_object( $license ) && FS_Plugin_License::is_valid_id( $license->parent_license_id ) );
-    $bundle_subscription = null;
+    $has_bundle_license              = ( is_object( $license ) && FS_Plugin_License::is_valid_id( $license->parent_license_id ) );
```
vovafeldman, line 166: @fajardoleo for improved readability, I'd suggest to change it to:
```
$has_bundle_license = (
    is_object( $license ) &&
    FS_Plugin_License::is_valid_id( $license->parent_license_id )
);
```

Also, I'm not sure about the new name you gave to that variable as it doesn't check if there's a bundle license, but checks if the context license is a child of a bundle license.

templates/account.php
```diff
@@ -163,26 +163,55 @@
...
 
     $active_plugins_directories_map = Freemius::get_active_plugins_directories_map( $fs_blog_id );
 
     $is_premium = $fs->is_premium();
+
+    $account_addons       = $fs->get_updated_account_addons();
+    $installed_addons     = $fs->get_installed_addons();
+    $installed_addons_ids = array();
+
+    foreach ( $installed_addons as $fs_addon ) {
```
vovafeldman, line 192: @fajardoleo any logic that is not trivial or longer than a few lines of code needs to be documented.

templates/account.php
vovafeldman: Actually, now that I reviewed the code below, I better understand the decision about the variable name. Therefore, a better way to write that code from a semantical point of view is as follows:
```
$has_bundle_license = false;

if ( is_object( $license ) &&
     FS_Plugin_License::is_valid_id( $license->parent_license_id )
) {
    // Context license has a parent license, therefore, the account has a bundle license.
    $has_bundle_license = true;
}
```

templates/account.php
```diff
@@ -163,26 +163,55 @@
...
+        $installed_addons_ids[] = $fs_addon->get_id();
+
+        if (
+            ! $has_bundle_license &&
+            $show_plan_row &&
+            $fs_addon->has_active_valid_license()
+        ) {
+            $addon_license = $fs_addon->_get_license();
+
+            $has_bundle_license = FS_Plugin_License::is_valid_id( $addon_license->parent_license_id );
```
vovafeldman, line 202: @fajardoleo same as the comment above, semantically it should be:
```
if ( FS_Plugin_License::is_valid_id( $addon_license->parent_license_id ) ) {
    // Add-on's license associated with a parent/bundle license.
    $has_bundle_license = true;
}

Comment (vovafeldman) [reply templates/account.php:202]: @fajardoleo also, the code in the `if` scope below can be merged with the logic of the proposed `if`.

File: templates/account.php [hunk 9]
```diff
@@ -163,26 +163,55 @@
...
+
+    $account_addons       = $fs->get_updated_account_addons();
+    $installed_addons     = $fs->get_installed_addons();
+    $installed_addons_ids = array();
+
+    foreach ( $installed_addons as $fs_addon ) {
+        $installed_addons_ids[] = $fs_addon->get_id();
+
+        if (
+            ! $has_bundle_license &&
```
vovafeldman, line 196: @fajardoleo since `$has_bundle_license` is the exceptional variable in this `if`, I would suggest to take it out from here and add the following code above the `if`:
```
if ( $has_bundle_license ) {
    // We already have the context bundle license details, skip.
    continue;
}
```

includes/class-freemius.php
fajardoleo: @vovafeldman `$blog_id` is new to this branch, so it's always `null` in old usages.

PR #366, opened 2019-09-01
developer-license / white-labeling

fajardoleo:


silvercolt:
Is there an explanation of this feature and it's intended use case(s)?

vovafeldman:
@silvercolt this feature was requested by agencies & freelancers that are building sites for others. It will allow them to flag a license as _white-labeled_, which will automatically set the Freemius SDK to run in a _white-labeled mode_ for all sites that are utilizing the license. It will hide things like the license owner information, license key, pricing page, etc.

PR #368, opened 2019-09-29
Feature/cloned db entities fix

fajardoleo:


includes/class-freemius.php
```diff
@@ -9985,6 +9985,115 @@ private static function set_account_option_by_module(
...
+            // Get first element.
+            $first_array_element = reset( $entities );
+
+            if ( $first_array_element instanceof $class_name ) {
+                /**
+                 * If the first element of the array is an instance of the context class, assume that all other
+                 * elements are instances of the class.
+                 */
+                return $entities;
+            } else if (
```
vovafeldman: @fajardoleo there is no reason for the `else if` since the prev `if` ends with a `return`

includes/class-freemius.php
```diff
@@ -9985,6 +9985,115 @@ private static function set_account_option_by_module(
...
+                is_array( $first_array_element ) &&
+                ! empty( $first_array_element ) &&
+                reset( $first_array_element ) instanceof $class_name
```
vovafeldman: @fajardoleo I'm not sure if PHP 5.3 will be happy with that syntax, I'd suggest setting a helper variable for `reset( $first_array_element )` within this `if` and then have another `if` inside.

vovafeldman:
@fajardoleo one final comment about the code, instead of hard coding the class names as strings, I suggest introducing a public static helper method to the core entity class that will look like that:
```
public static function get_class_name() {
    return get_called_class();
}
```

And then you can pass the class name as `FS_Subscription::get_class_name()`. It is a better practice for the long run, in case we'd choose to refactor any of those class names.

PR #382, opened 2019-12-26
[plugin-actions] [js] The JS handlers for the plugin action links wer…

fajardoleo:
…e not working with the search results content (on Plugins and Themes pages).

### DIFF
--- a/templates/forms/deactivation/form.php
+++ b/templates/forms/deactivation/form.php
@@ -79,8 +79,8 @@ HTML;
 ?>
 <script type="text/javascript">
 (function ($) {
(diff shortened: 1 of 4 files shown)


dbjpanda:
@fajardoleo  Thanks for the email. I will let you know soon if it fixes my issue.

templates/forms/deactivation/form.php
```diff
@@ -181,7 +181,11 @@
 	registerEventHandlers();
 
 	function registerEventHandlers() {
-		$deactivateLink.click(function (evt) {
+		$body.on( 'click', '#the-list .deactivate > a', function ( evt ) {
```
dbjpanda: This line actually fixes the real issue. :)

PR #390, opened 2020-03-03
Set the max width for iframes (videos) to 100%

dovy:
If not set, the videos from descriptions bleed in the dialog box. 

See


vovafeldman:
@dovy thanks for the patch!

PR #393, opened 2020-03-18
Do not override WordPress.org plugin descriptions inheritance

sc0ttkclark:
This PR will prevent the empty Freemius add-on description from overriding the WordPress.org plugin description and adds a new filter to customize the description further.

### DIFF
--- a/includes/fs-plugin-info-dialog.php
+++ b/includes/fs-plugin-info-dialog.php
@@ -328,7 +328,15 @@
                 }
             }
 
-            $data->sections['description'] = fs_get_template( '/plugin-info/description.php', $view_vars );
+            $description = fs_get_template( '/plugin-info/description.php', $view_vars );
+            $description = trim( $description );
+
+            $description = $this->_fs->apply_filters( 'plugin_section_description', $description, $data );
+
+            // Only override description if it's not empty or not set yet.
+            if ( '' !== $description || ! isset( $data->sections['description'] ) ) {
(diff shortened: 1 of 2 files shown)


templates/plugin-info/description.php
```diff
@@ -32,7 +33,11 @@
 				<?php endfor ?>
 			</ul>
 		</div>
-	<?php endif ?>
+	<?php
+	endif;
+	
+	if ( ! empty( $plugin->info->description ) ) :
```
sc0ttkclark: This new logic prevents an empty `<div>      </div>` from appearing as the description when there's no description to show.

includes/fs-plugin-info-dialog.php
```diff
@@ -328,7 +328,15 @@ function _get_addon_info_filter( $data, $action = '', $args = null ) {
...
 
-            $data->sections['description'] = fs_get_template( '/plugin-info/description.php', $view_vars );
+            $description = fs_get_template( '/plugin-info/description.php', $view_vars );
+            $description = trim( $description );
+
+            $description = $this->_fs->apply_filters( 'plugin_section_description', $description, $data );
+
+            // Only override description if it's not empty or not set yet.
+            if ( '' !== $description || ! isset( $data->sections['description'] ) ) {
+                $data->sections['description'] = $description;
```
sc0ttkclark: This logic will allow the WordPress.org plugin description to not be replaced with an empty Freemius add-on description.

PR #397, opened 2020-03-31
Fix array_splice bug removing wrong plugin from active_plugins

sc0ttkclark:
Debugging and props really belong to @nobleclem via https://github.com/pods-framework/pods/issues/5646

### DIFF
--- a/includes/fs-essential-functions.php
+++ b/includes/fs-essential-functions.php
@@ -387,7 +387,7 @@
 
         $newest_sdk_plugin_path = $fs_active_plugins->newest->plugin_path;
 
-        $active_plugins        = get_option( 'active_plugins', array() );
+        $active_plugins        = array_values( get_option( 'active_plugins', array() ) );
         $newest_sdk_plugin_key = array_search( $newest_sdk_plugin_path, $active_plugins );
         if ( 0 === $newest_sdk_plugin_key ) {
             // if it's 0 it's the first plugin already, no need to continue
--- a/includes/fs-essential-functions.php
+++ b/includes/fs-essential-functions.php
@@ -393,8 +393,9 @@
             // if it's 0 it's the first plugin already, no need to continue
             return false;
         } else if ( is_numeric( $newest_sdk_plugin_key ) ) {
-            // Remove plugin from its current position (position minus one for array_splice usage).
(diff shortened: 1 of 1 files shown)


includes/fs-essential-functions.php
```diff
@@ -387,14 +387,14 @@ function fs_newest_sdk_plugin_first() {
...
+        $active_plugins        = array_values( get_option( 'active_plugins', array() ) );
         $newest_sdk_plugin_key = array_search( $newest_sdk_plugin_path, $active_plugins );
         if ( 0 === $newest_sdk_plugin_key ) {
             // if it's 0 it's the first plugin already, no need to continue
             return false;
         } else if ( is_numeric( $newest_sdk_plugin_key ) ) {
-            // Remove plugin from its current position.
-            array_splice( $active_plugins, $newest_sdk_plugin_key, 1 );
+            // Remove plugin from its current position (position minus one for array_splice usage).
+            array_splice( $active_plugins, $newest_sdk_plugin_key - 1, 1 );
```
sc0ttkclark:
```suggestion
            // Remove plugin from its current position.
            array_splice( $active_plugins, $newest_sdk_plugin_key, 1 );

```

vovafeldman:
It's the first time we are seeing this issue. I wonder how the indexes got mixed up on @nobleclem's environment from the first place. Regardless, the addition of `array_values()` should do the trick. That said, it is not the most performant solution so, I pushed the following fix https://github.com/Freemius/wordpress-sdk/pull/398

Please give it a try (**please use a test environment**) and let me know if it resolved the issue.

nobleclem:
@vovafeldman I am not sure how the index keys got broken.  This is on a multisite install and not every site on the install has this active_plugins missing sequence issue.

One thing to note that I realized when working with @sc0ttkclark on this was that this code only runs on install I think.  And after it runs its reorder that it doesn't run again.  So if I activate another plugin the order of active_plugins returns back to its default of alpha sort.  So then any plugin using freemius is no longer the first in the list.

1) activate plugin using freemius
2) check the value of active_plugins (freemius plugin should be at top of list as expected)
3) activate a non freemius plugin (the plugin that was in the first position in step 2 is no longer there and the active_plugins list will be in alpha order)

sc0ttkclark:
Closing this in favor of #398

PR #398, opened 2020-04-01
Feature/plugins reorder fix

vovafeldman:
Optimization of https://github.com/Freemius/wordpress-sdk/pull/397

### DIFF
--- a/includes/fs-essential-functions.php
+++ b/includes/fs-essential-functions.php
@@ -417,23 +417,25 @@
             update_option( 'active_plugins', $updated_active_plugins );
 
             return true;
-        } else if ( is_multisite() && false === $newest_sdk_plugin_key ) {
+        }
+
+        if ( is_multisite() ) {
             // Plugin is network active.
             $network_active_plugins = get_site_option( 'active_sitewide_plugins', array() );
 
-            if (isset($network_active_plugins[$newest_sdk_plugin_path])) {
-                reset($network_active_plugins);
-                if ( $newest_sdk_plugin_path === key($network_active_plugins) ) {
+            if ( isset( $network_active_plugins[ $newest_sdk_plugin_path ] ) ) {
+                reset( $network_active_plugins );
+                if ( $newest_sdk_plugin_path === key( $network_active_plugins ) ) {
                     // Plugin is already activated first on the network level.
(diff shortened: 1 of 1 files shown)


nobleclem:
It looks like the usage of `$is_first_path` is broken in this PR.

It's initially set to true then there is a check that if its false set it to false on L411.  But in the foreach loop on L396 there is a check if its true (default) return false.  So basically the first interation through the loop it would return false no?

vovafeldman:
Good catch, @nobleclem :) Committed the fix.

nobleclem:
So this looks like it works well to fix the ordering bug.  Just a note though on how a numeric index went missing while testing it happens when you deactivate a plugin.  So if you have say 4 plugins active and you remove the second in the list then your indexes would be 0,2,3.  I have not looked but I suspect that wordpress just simply unset()'s the plugin from the array.

However the issue now of when you activate a non-freemius plugin after this reorder has already happened still exists.  So potentially older freemius libraries will load instead of the newest one.

vovafeldman:
@nobleclem that is very strange - the SDK is running on millions of sites, and it's the first time someone raised that issue (in ~5 years).

Regarding your concern about older SDK versions - that is not a problem. The whole purpose of that mechanism is to automatically position the plugin that has the newest SDK version first (so it will be included first by WP core). Since the fixed version will have a newer SDK version, the SDK will not trigger the reorder logic.

nobleclem:
So I understand your reluctance on the ordering issue.  People need to see issues in order to report them which is why I am here.  It just happens when debugging and testing I found a secondary issue with the order change this SDK perform and how it can easily be reset back to WP default order by simply activating a plugin that doesn't contain this SDK.  So lets take the scenario of I have two plugins using this SDK (Justified Gallery and Pods Framework).  The default WP ordering would have Justified Gallery load which would use its version of the SDK.  BUT with this reordering code it actually switches that to putting Pods first because it has a newer SDK.  But what happens when I activate another plugin later.  In that lets say I activate "BU Navigation".  Well when I look at the active_plugins value its back to the WP default order which would be: BU, Justified, and Pods.  So wouldn't that mean the SDK of Justified is loading and not the newer version that is contain in Pods?

PR #399, opened 2020-04-02
[opt-in] Fixed the new install identification.

fajardoleo:


vovafeldman:
The logic doesn't cover the following use-case:

1. User install pluginX before it was integrated with Freemius
2. Then, they deactivate the plugin
3. Update the plugin to a newer version integrated with Freemius
4. Activates the plugin back

In this case, the logic will consider the plugin as a new install and not an update.

Also, seems like the complimentary theme new install identification logic is no longer working, something we'll need to fix in the future.

**Notes from @fajardoleo:**

This is what happens when I activate a Freemius-powered plugin:

1. The “Activate” link on the Plugins page is clicked
2. That will load `wp-config.php` and then `wp-settings.php`.
3. `wp-settings.php` will eventually load active plugins (excluding the current plugin that is being activated)
4. `wp-settings.php` will execute `do_action( 'plugins_loaded' )`.
5. And then the activated plugin’s main file will eventually be included, so in this case, `did_action( 'plugins_loaded' )` will return `1`

PR #426, opened 2020-06-24
[bundle] [license] Handled the case when the core plugin is network-a…

fajardoleo:
…ctive and activated with a bundle license, and an add-on is site-level–activated and the bundle license needs to be activated for the add-on.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -4941,6 +4941,12 @@
                             $premium_license = null;
 
                             if ( ! $this->has_free_plan() && $this->_parent->has_features_enabled_bundle_license() ) {
+                                /**
+                                 * If the add-on has no free plan, try to activate with a bundle license if there's any.
+                                 *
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -10634,6 +10678,18 @@ function is_tracking_allowed() {
...
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since  2.3.3
+         *
+         * @return bool
+         */
+        function is_bundle_license_auto_activation_enabled() {
+            return $this->is_addon() ?
+                $this->_parent->is_bundle_license_auto_activation_enabled() :
```
vovafeldman, line 10689: @fajardoleo I'd recommend to tweak it to:
`( is_object( $this->_parent ) && $this->_parent->is_bundle_license_auto_activation_enabled() )`

Just in case that this method is called from an add-on before the parent instance was initialized.

Issue #422, opened 2020-06-15
Fix the "_blank" link target vulnerability


dgwyer:
# 🐞 bug report

**Behavior**:
 -  `What is the issue?`
Link _blank target exploitation.

 -  `What is the expected behavior?`
Apply the fix as suggested here: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

**Versions**: 
Fix in latest Freemius SDK version.

dgwyer:
I don't mind taking this one.

dgwyer:
This has now been fixed and merged into the `develop` branch.

PR #431, opened 2020-06-29
Feature/422 fix blank target

dgwyer:
Fixed all `target="_blank"` attribute vulnerabilities by adding an additional `rel="noopener noreferrer"` attribute.


dgwyer:
Fixes #422.

PR #447, opened 2020-09-18
added wp_delete_site

Dudo1985:
Since version 5.1.0 deleted_blog is deprecated.

I've added a control, and if wp version is > 5.10 then use wp_delete_site instead of deleted_blog

More info can be found here https://rudrastyh.com/wordpress-multisite/delete-blog-deprecated.html


vovafeldman:
@Dudo1985 thanks for the catch! Question: why did you write a new method instead of reusing `_after_site_deleted_callback()`?

Dudo1985:
I wasn't sure about this: then, I thought the soon or later support for `deleted_blog` will come to an end, and will be easier then to just remove the method `_after_site_deleted_callback` instead of editing it again

asadowski10:
Hello,

We have several premium plugins using the SDK and today we are stuck on removing sites.

Can you prioritize this ticket so that we stop hacking premium plugins?

thank you,

vovafeldman:
Note: The action `wpmu_new_blog` was also deprecated and we need to update the code to use `wp_insert_site`.

PR #448, opened 2020-09-24
Verify PHP 7.x compatibility

stevegrunwell:
While working with some compatibility checks, I noticed that a number of plugins that include Freemius' SDK were being flagged as having compatibility issues with PHP 7.x.

Most of the issues were warnings about functions that inspect arguments (such as `func_get_args()` and `debug_backtrace()`), as PHP 7.x will report the **current** values of those variables rather than what was originally passed:

> Since PHP 7.0, functions inspecting arguments, like func_get_args(), no longer report the original value as passed to a parameter, but will instead provide the current value. ([source](https://github.com/PHPCompatibility/PHPCompatibility/blob/develop/PHPCompatibility/Sniffs/FunctionUse/ArgumentFunctionsReportCurrentValueSniff.php#L21))

In one case, `debug_backtrace()` is being made more efficient by ignoring the arguments (thus resolving the warning) and limiting the stack size, as the backtrace is only being used to find the caller.

The one explicit **error** revolves around `idn_to_ascii()`, as [the default `$variant` argument has changed over time](https://www.php.net/manual/en/function.idn-to-ascii.php#refsect1-function.idn-to-ascii-changelog); to ensure maximum compatibility, the variant has been set to `INTL_IDNA_VARIANT_UTS46`, available since PHP 5.4 and the default as of PHP 7.4.

This PR includes [automated compatibility checks via phpcompatibility/phpcompatibility](https://github.com/PHPCompatibility/PHPCompatibility), the aforementioned fixes, as well as a GitHub Actions workflow for automatically executing the checks on push and/or PR.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,0 +1,29 @@
+name: Coding standards
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
+    - name: Cache Composer packages
+      id: composer-cache
+      uses: actions/cache@v2
+      with:
+        path: vendor
+        key: ${{ runner.os }}-node-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+
+    - name: Install dependencies
+      if: steps.composer-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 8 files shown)


vovafeldman:
Thanks a lot for your contribution, @stevegrunwell 🙏

bahiirwa:
@stevegrunwell Thank you for this elaborate PR.

Your efforts highlighted the need and have allowed us to build the CI further in https://github.com/Freemius/wordpress-sdk/pull/646 to have
- [x] static analysis using PHPStan 
- [x] PHPCompatibility checks

In light of merging #646, we shall close this ticket.

PR #465, opened 2020-12-28
[whitelabel] [filter] Added a filter for the whitelabeled mode flag.

fajardoleo:


vovafeldman:
@fajardoleo  why not simply change this line:

`if ( true !== $this->_storage->is_whitelabeled ) {`

with:

`if ( true !== $this->apply_filters( 'is_whitelabeled', $this->_storage->is_whitelabeled ) ) {`

?

PR #486, opened 2021-03-24
Feature/admin notices about page fix

xiahengchen:
[admin-notices] [about-page] [bug-fix] Hide the admin notices on the 'About' page.


vovafeldman:
@xiahengchen this is a wrong implementation as it will remove ALL admin notices, including notices that are not generated by the SDK. Please check [this commit](https://github.com/Freemius/wordpress-sdk/commit/8a87d389c647b4588bfe96fc7d420d62a48cbac5) in which we stopped showing notices on Gutenberg pages. I would suggest adding a helper function to determine if notices should be shown or not that will check for the about page + gutenberg.

PR #489, opened 2021-04-19
add to composer autoload

afragen:
This PR adds the `start.php` file to composer's autoload function if the SDK is installed via composer. This simplifies the initiation of the SDK by not requiring the use of the `require`.


afragen:
This will cause issues with Bedrock.

PR #497, opened 2021-05-24
[menu] [menu-url] Fixed an issue with the menu URL generation logic.

fajardoleo:


vovafeldman:
Related to [this ticket](https://secure.helpscout.net/conversation/1518645188/15195/#thread-4421452558), which no longer needs the fix.

I'm closing this one since it may trigger unexpected issues. We can reopen it in the future based on demand.

PR #514, opened 2021-08-08
Feature/clone resolution

fajardoleo:


templates/clone-resolution-js.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+?>
+<script type="text/javascript">
+    ( function( $ ) {
+        var $errorMessage = null;
+
+        $( document ).ready( function() {
+            var $cloneResolutionNotice = $( 'div[data-id="clone_resolution_options_notice"], div[data-id="temporary_duplicate_notice"]' );
+
+            if ( 1 === $cloneResolutionNotice.length ) {
```
vovafeldman, line 20: @fajardoleo cleaner to use:
```
if ( 0 === $cloneResolutionNotice.length ) {
    return;
}
```

and then have the rest of the code.

templates/clone-resolution-js.php
```diff
@@ -0,0 +1,75 @@
...
+                $errorMessage = $cloneResolutionNotice.find( '#fs_clone_resolution_error_message' );
+
+                $cloneResolutionNotice.on( 'click', '.button, #fs_temporary_duplicate_license_activation_link', function( evt ) {
```
vovafeldman, line 23: @fajardoleo missing documentation.

config.php
```diff
@@ -251,6 +251,9 @@
     if ( ! defined( 'WP_FS__GDPR_OPTION_NAME' ) ) {
         define( 'WP_FS__GDPR_OPTION_NAME', WP_FS___OPTION_PREFIX . 'gdpr' );
     }
+    if ( ! defined( 'WP_FS__CLONE_MANAGEMENT_OPTION_NAME' ) ) {
```
vovafeldman, line 254: @fajardoleo the define is only used in the clone manager, and only once, so I think there's no need to have a "define" for it.

includes/managers/class-fs-clone-manager.php
```diff
@@ -0,0 +1,421 @@
...
+            }
+
+            return self::$_instance;
+        }
+
+        #endregion
+
+        private function __construct() {
+            $this->_storage     = FS_Option_Manager::get_manager( WP_FS__CLONE_MANAGEMENT_OPTION_NAME, true );
+            $this->_option_name = 'clone_resolution';
```
vovafeldman, line 61: @fajardoleo seems like it's a constant. If that's the case, let's make it one.

includes/managers/class-fs-clone-manager.php
```diff
@@ -0,0 +1,421 @@
...
+                add_action( 'admin_post_fs_clone_resolution', array( &$this, '_handle_clone_resolution' ) );
+
+                return;
+            }
+
+            $this->initiate_clone_resolution_handler();
+        }
+
+        /**
+         * @return int|null
```
vovafeldman, line 81: @fajardoleo please add an explanation, even if short and relatively easy.

includes/managers/class-fs-clone-manager.php
```diff
@@ -0,0 +1,421 @@
...
+     * @license   https://www.gnu.org/licenses/gpl-3.0.html GNU General Public License Version 3
+     * @author    Leo Fajardo (@leorw)
+     * @since     2.4.3
+     */
+
+    if ( ! defined( 'ABSPATH' ) ) {
+        exit;
+    }
+
+    class FS_Clone_Manager {
```
vovafeldman, line 14: @fajardoleo Add PHP Doc + explain the purpose of the class

includes/managers/class-fs-clone-manager.php
```diff
@@ -0,0 +1,421 @@
...
+        private $_storage;
+        /**
+         * @var array {
+         * @type bool   $is_temporary_duplicate
+         * @type int    $clone_identification_timestamp
+         * @type int    $request_handler_timestamp
+         * @type string $request_handler_id
+         * }
+         */
+        private $_data;
```
vovafeldman, line 27: @fajardoleo instead of accessing the data through `$this->_data['key']`, I suggest to enrich the class with the [magic methods](https://www.php.net/manual/en/language.oop5.overloading.php#object.get) `__get()`, `__set()`, `__isset()` (based on what's needed), and add the relevant properties as `@property`-ies in the PHP doc of the class for auto-complete. This way the code will be more structured (we can use auto-complete, refactoring, etc.)

includes/class-freemius.php
```diff
@@ -3514,6 +3519,411 @@ private static function _load_required_static() {
...
+            $clone_action = fs_request_get( 'clone_action' );
+
+            if ( empty( $clone_action ) ) {
+                self::shoot_ajax_failure( array(
+                    'message'      => fs_text_inline( 'Invalid clone resolution action param.', 'invalid-clone-resolution-action-param-error' ),
+                    'redirect_url' => '',
+                ) );
+            }
+
+            if ( 'temporary_duplicate' === $clone_action ) {
```
vovafeldman, line 3596: @fajardoleo clone actions should probably be constants in the clone manager instead of strings

includes/class-freemius.php
```diff
@@ -3514,6 +3519,411 @@ private static function _load_required_static() {
...
+            );
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.4.3
+         * 
+         * @return string
+         */
+        private static function get_temporary_duplicate_admin_notice_string( $site_urls, $module_label ) {
```
vovafeldman, line 3883: @fajardoleo this can go into the clone manager

includes/class-freemius.php
```diff
@@ -5273,7 +5693,7 @@ function stop_site_tracking() {
             }
 
             // Send update to FS.
-            $result = $this->get_api_site_scope()->call( '/?fields=is_disconnected', 'put', array(
+            $result = $this->get_api_site_scope()->call( $this->add_is_clone_param( '/?fields=is_disconnected' ), 'put', array(
```
vovafeldman, line 5696: @fajardoleo instead explicitly adding `$this->add_is_clone_param()`, from a code design perspective, it's better to create a helper method(s) that will wrap `$this->get_api_site_scope()->call()` and `$this->get_api_site_scope()->get()` and use those. That way, we won't need to remember it when adding new calls in the future + if we would want to change something in the logic that maybe impacts more than just the querystring param, we'll be able to make a change in one (or two) places.

includes/class-freemius.php
```diff
@@ -10919,6 +11387,50 @@ function get_site() {
...
+        private function delete_current_install( $back_up ) {
+            // Back up and delete the unique ID.
+            if ( $back_up ) {
+                self::$_accounts->set_option( 'prev_unique_id', $this->get_anonymous_id() );
+            }
+
+            self::$_accounts->set_option( 'unique_id', null );
+
+            if ( $back_up ) {
+                // Back up and delete the current install.
```
vovafeldman, line 11412: @fajardoleo it's not clear why you need to "backup" the site, isn't the site was loaded from the DB already? why would we need to store it again. Also, the comment is incorrect, this line doesn't delete the install.

includes/class-freemius.php
```diff
@@ -10037,7 +10495,12 @@ public static function _uninstall_plugin_hook() {
                     return;
                 }
 
-                $fs->_uninstall_plugin_event();
+                if (
+                    ! $fs->is_clone() &&
+                    ( ! is_object( $fs->_site ) || $fs->is_registered() )
```
vovafeldman, line 10500: @fajardoleo it isn't clear why specifically when there's no install context (`! is_object( $fs->_site )`) we should run the uninstall handler.

includes/class-freemius.php
```diff
@@ -23112,13 +23701,69 @@ private function get_api_site_scope( $flush = false ) {
...
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.4.3
+         *
+         * @param array|mixed|string|void $api_result
+         * @param bool                    $check_if_entity
+         */
+        private function maybe_update_local_install( $api_result ) {
```
vovafeldman, line 23739: @fajardoleo it's not clear what's the purpose of this method, please document.

includes/class-fs-api.php
```diff
@@ -198,6 +211,17 @@ private function _call( $path, $method = 'GET', $params = array(), $retry = fals
                     }
                 }
 
+                /**
+                 * @since 2.4.3
```
vovafeldman, line 215: @fajardoleo please add short explanation

includes/class-freemius.php
fajardoleo: @vovafeldman We back up a site so that we can restore it if the automatic resolution attempt fails (during an automatic resolution attempt, we delete the site).

includes/class-freemius.php
fajardoleo: @vovafeldman When there's no install context, it works like before. On the other hand, if there's an install context, it should work like before only if there's a valid user entity.

includes/class-freemius.php
vovafeldman: @fajardoleo in that case please update the comment accordingly. Also, add documentation to the method that explains the potential backup and reasoning.

includes/class-freemius.php
vovafeldman: @fajardoleo `it should work like before only if there's a valid user entity` it is not clear to me why. Please add a comment that explains it and why.

PR #522, opened 2021-10-19
[clone-resolution] [notices] [fix] Do not show the clone resolution a…

fajardoleo:
…dmin notice if there is no Freemius instance that is associated with a clone/cloned site.

### DIFF
--- a/includes/class-fs-admin-notices.php
+++ b/includes/class-fs-admin-notices.php
@@ -128,11 +128,7 @@
             $network_level_or_blog_id = null,
             $is_dimissible = null
         ) {
-            if ( $this->should_use_network_notices( $id, $network_level_or_blog_id ) ) {
-                $notices = $this->_network_notices;
-            } else {
-                $notices = $this->get_site_notices( $network_level_or_blog_id );
-            }
+            $notices = $this->get_site_or_network_notices( $id, $network_level_or_blog_id );
 
             $notices->add(
                 $message,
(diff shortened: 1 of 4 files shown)


includes/managers/class-fs-clone-manager.php
```diff
@@ -694,17 +733,96 @@ private function maybe_show_clone_admin_notice() {
...
+                if ( empty( $notice['data'] ) || ! is_array( $notice['data'] ) ) {
+                    continue;
+                }
+
+                if ( empty( $notice['data']['product_ids'] ) || empty( $notice['data']['blog_id'] ) ) {
+                    continue;
+                }
+
+                $product_ids = $notice['data']['product_ids'];
+                $blog_id     = $notice['data']['blog_id'];
```
vovafeldman, line 767: @fajardoleo it's very much possible that a subsite will be deleted and therefore, `switch_to_blog()` will not function as expected. Thus, I suggest using [`get_site()`](https://github.com/WordPress/wordpress-develop/blob/5.8.1/src/wp-includes/ms-site.php#L308) first to check if the result isn't `null` and only then continue with the logic.

includes/class-fs-admin-notices.php
```diff
@@ -216,15 +217,37 @@ function add_sticky(
...
+         * @author Leo Fajardo (@leorw)
+         * @since 2.4.3
+         *
+         * @param string   $id
+         * @param int|null $network_level_or_blog_id
+         *
+         * @return array|null
+         */
+        function get_sticky( $id, $network_level_or_blog_id ) {
+            if ( $this->should_use_network_notices( $id, $network_level_or_blog_id ) ) {
```
vovafeldman, line 244: @fajardoleo this `if/else` for the right notices object selection is now showing 4 times in the same class. I suggested introducing `get_site_or_network_notices()` as a helper private method to reduce code redundancy.

PR #526, opened 2021-12-04
[bugfix] prevent fatal error on plugin activation when SECURE_AUTH_KEY is missing

alewolf:
This prevents a fatal error when a user tries to activate the plugin and the SECURE_AUTH_KEY is missing in wp-config.php


zocoloco:
I just purchased something that uses Freemius and got this same error. (PHP8). Has it not been fixed yet?

PR #527, opened 2021-12-09
Feature/clone resolution multisite related fixes

fajardoleo:


includes/class-freemius.php
```diff
@@ -3610,6 +3608,100 @@ private static function _load_required_static() {
         #region Clone
         #--------------------------------------------------------------------------------
 
+        /**
+         * Tries to recover the install of a newly created subsite or resolve it if it's a clone.
+         * 
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.0
+         */
+        private function maybe_resolve_new_subsite_install_automatically() {
```
vovafeldman, line 3617: @fajardoleo can we move this method, and the other newly created helper methods into the `FS_Clone_Manager`? That way we'll also be able to revert the visibility of the methods in `FS_Clone_Manager` that are used back to `private`.

vovafeldman (changes requested):
Well done, @fajardoleo 👏 
Only a few minor comments. Everything looks as we discussed, well commented, and structured.

includes/class-freemius.php
```diff
@@ -9862,37 +9934,29 @@ private function send_install_update( $override = array(), $flush = false ) {
...
             $this->_logger->entrance();
 
-            $installs_data = $this->get_installs_data_for_api( $override, ! $flush );
+            /**
+             * Pass `true` to use the network level storage since the update is for many installs.
+             *
+             * @author Leo Fajardo (@leorw)
+             * @since 2.2.3
+             */
+            $keepalive_only_update = $this->should_send_keepalive_update( true );
```
vovafeldman, line 9950: @fajardoleo the purpose/meaning of this param has changed now and it is no longer indicating whether it's only a keepalive update or not, so I propose to refactor to `$should_send_keepalive`.

templates/account/partials/site.php
```diff
@@ -174,7 +174,7 @@ class="dashicons dashicons-arrow-right-alt2"></i>
                 <?php $row_index ++ ?>
                 <!--/ Blog ID -->
 
-                <?php if ( $is_registered && $install->user_id != $main_user->id ) : ?>
+                <?php if ( $install->user_id != $main_user->id ) : ?>
```
vovafeldman, line 177: @fajardoleo don't we want to check `is_object( $VARS['user'] )` as the replacement for `$is_registered`?

templates/forms/license-activation.php
```diff
@@ -59,13 +59,60 @@
...
+                }
+            }
 
+            $site_info_by_install_id_map[] = array(
+                'site'    => $site,
+                'install' => $install,
+            );
+        }
+
+        foreach ( $site_info_by_install_id_map as $site_info )
```
vovafeldman, line 91: @fajardoleo similar to one of the previous comments in `get_installs_data_for_api()`, what's the purpose of the additional `foreach`? why not handle the logic within the first `foreach`?

vovafeldman (changes requested):

includes/class-freemius.php
```diff
@@ -9659,59 +9697,93 @@ private function get_installs_data_for_api(
...
+                $site_info_by_install_id_map[ $install->id ] = array(
+                    'site'    => $site,
+                    'install' => $install,
+                );
+            }
 
-                    $install_data['is_disconnected'] = $install->is_disconnected;
-                    $install_data['is_active']       = $this->is_active_for_site( $blog_id );
-                    $install_data['is_uninstalled']  = $install->is_uninstalled;
+            foreach ( $site_info_by_install_id_map as $site_info ) {
```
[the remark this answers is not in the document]
fajardoleo: @vovafeldman Its purpose is to prevent executing `$install = $this->get_install_by_blog_id( $blog_id );` more than once in case of duplicates since it can load the options table of more than one blog only to find the same install. The other purpose is to avoid removing entries from the `$installs_data` array if later on, we find a duplicate.

templates/forms/license-activation.php
fajardoleo: @vovafeldman The main purpose is to prevent removing entries from the `$sites_details` array if later on, we find a subsite with a clone install.

templates/account/partials/site.php
fajardoleo: @vovafeldman The `$is_registered` part was removed as the modified line is already being wrapped with `if ( $is_registered )...`. With regard to `$VARS['user']`, the "Account" page is loaded only when it's an object.

includes/class-freemius.php
fajardoleo: @vovafeldman Renamed, thanks.

Issue #453, opened 2020-10-15
Some Freemius pages are not mobile friendly


iksent:
# 🐞 bug report

**Behavior**:
1. Account page is really bad at mobile devices:

![2020-10-15_19-20-30](https://user-images.githubusercontent.com/32956521/96142708-d363d180-0f1b-11eb-95b2-d6246b47fa19.png)
![2020-10-15_19-20-54](https://user-images.githubusercontent.com/32956521/96142705-d19a0e00-0f1b-11eb-8e15-fd6286ae07ab.png)

2. Texts at header of "Contact us" page are wraping at the new line (from 551px to 648px)

![2020-10-15_19-21-47](https://user-images.githubusercontent.com/32956521/96142782-e5de0b00-0f1b-11eb-9075-ddf2a81ff564.png)

**Versions**: (*)
 -  `Freemius SDK Version:` 2.4.0
 -  `WordPress Version:` 5.5.1
 -  `PHP Version:` 7.0

**Additional Information**:
 -  `Browser Type: Latest Chrome`
 -  `OS: Windows`

PR #528, opened 2021-12-10
[bug-fix]: make account page look better on mobile devices

swashata:
This fixes the Accounts Page issue from #453. I tried to reproduce the Contact Us page issue, but wasn't able to do so. I believe it was fixed through some other release?

### DIFF
--- a/assets/scss/admin/account.scss
+++ b/assets/scss/admin/account.scss
@@ -316,3 +316,63 @@ label.fs-tag
         width: 100%;
     }
 }
+
+// Fixes for issue 453.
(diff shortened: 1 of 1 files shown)


swashata:
Also kindly sqash the two commits into one, the first one is where I couldn't figure out how to compile the sass files using gulp. Then I realized it wasn't there, so I just went for globally installed sass CLI.

swashata:
Please consider the last commit, if you actually end up pulling this.

swashata:
I have fixed the raised issue on the PR and have squashed the previous commits to keep a clean history on the develop branch. Please review.

PR #529, opened 2021-12-19
[admin-notice] [data-sync] Implemented the logic to show relevant admin notice only after 3 failed attempts for background process.

msladu:


includes/class-freemius.php
```diff
@@ -20980,8 +20980,8 @@ private function _sync_plugin_license(
                         $this->switch_to_blog( $this->_storage->network_install_blog_id );
                     }
 
-                    // Show API messages only if not background sync or if paying customer.
-                    if ( ! $background || $this->is_paying() ) {
+                    // Show API messages only if paying customer.
+                    if ( $this->is_paying() ) {
```
vovafeldman, line 20984: @msladu when there's an API connectivity issue for a non-background request we **do** want to show the error notice, regardless of the `$this->is_paying()` result. So unless I'm missing something, this change should be reverted.

includes/class-freemius.php
```diff
@@ -20992,22 +20992,38 @@ private function _sync_plugin_license(
...
-                                            'wp.freemius.com'
-                                        ) ) ) . '</b>'
-                                    ) . '<br> ' . $this->get_text_inline( 'Error received from the server:', 'server-error-message' ) . var_export( $result->error, true ),
-                                    $this->get_text_x_inline( 'Oops', 'exclamation', 'oops' ) . '...',
-                                    'error',
-                                    $background,
-                                    'api_blocked'
-                                );
-                            }
+	                        $add_notice = false;
```
vovafeldman, line 20995: @msladu the entire code from this line and below doesn't seem to be properly indented.

includes/class-freemius.php
```diff
@@ -20992,22 +20992,38 @@ private function _sync_plugin_license(
...
+	                        if ( $background ) {
+		                        $counter = (int) get_transient( '_fs_api_connection_retry_counter' );
+
+		                        if ( $counter >= 3 ) {
+			                        $add_notice = true;
+			                        $counter    = 0;
```
vovafeldman, line 21001: @msladu 
1. Resetting the counter after 3 failures is a good idea, but it's not obvious and requires an explanation that we only want to show the notice after 3 consecutive failures, and then if the notice was already shown, start the counter from scratch.
2. In addition to resetting it here, we also need to reset the counter (actually delete the transient) when the API request was valid. Otherwise, if the API fails twice, then succeeds, and then fails one more time, the message will be shown, even though there was only one consecutive connectivity failure and not 3.

includes/class-freemius.php
```diff
@@ -20992,22 +20992,38 @@ private function _sync_plugin_license(
...
+		                        }
+
+                                // Generally, background sync happens once in a day. So, setting expiration for a week time is fair enough.
```
vovafeldman, line 21004: @msladu grammar notes :)
-  `... once a day ...` not `... once in a day ...`. 
- the word `fair` is irrelevant here, `... a week time should be enough.`

includes/class-freemius.php
msladu: @vovafeldman You are right. 

So far, we are showing admin notice only for non background process or for paying customer. 

Now, we want to do API test for non background process also. Do we want to check only for paying customer or also want to check for non paying customer?

includes/class-freemius.php
msladu: @vovafeldman  

1. Yes, I will add the explanation. 
2. Got it. Will implement that.

includes/class-freemius.php
msladu: @vovafeldman It looks proper in my IDE setup.

includes/class-freemius.php
msladu: @vovafeldman Ok.

includes/class-freemius.php
vovafeldman: I'm not sure I get your question, so perhaps I'll try to clarify all the cases:

* If it's a non-background sync request (i.e., `$background = false`), if there's a connectivity issue => show error notice immediately whether `is_paying()` is `true` or `false` (no retries mechanism or something like that, the user needs to see an error since they clicked on something that supposed to do something with the API).
* If it's a background request (i.e., `$background = true`) and `is_paying()` returns `false` (i.e., the user is not on a paid plan so even if API requests fail, there's no reason to bother the user since it's not mission-critical for the functionality of the product).
* If it's a background request  (i.e., `$background = true`) and `is_paying()` returns `true` that's where we want to show the error notice (since lack of connectivity can cause issues). But, this is also the case where we want to add the retries mechanism so in case it's a temporary connectivity issue we won't bother the user about it.

Hope this is clearer.

msladu: Thanks @vovafeldman!

It clarifies!

includes/class-freemius.php
```diff
@@ -20980,7 +20980,7 @@ private function _sync_plugin_license(
                         $this->switch_to_blog( $this->_storage->network_install_blog_id );
                     }
 
-                    // Show API messages only if not background sync or if paying customer.
+	                // Show API message only if not background sync or if paying customer. And if it's a background process, show message only if it's a paying customer.
```
vovafeldman, line 20983: @msladu The extra clarification you have added is redundant, please remove it.

includes/class-freemius.php
```diff
@@ -20992,23 +20992,43 @@ private function _sync_plugin_license(
...
-                                    $background,
-                                    'api_blocked'
-                                );
-                            }
+	                        $add_notice = false;
+	                        if ( $background ) {
+		                        $counter = (int) get_transient( '_fs_api_connection_retry_counter' );
+
+                                // We only want to show the notice after 3 consecutive failures.
+		                        if ( $counter >= 3 ) {
```
vovafeldman, line 21000: @msladu please revert the order of the arguments, our convention is to put the constant on the left side.

includes/class-freemius.php
```diff
@@ -20992,23 +20992,43 @@ private function _sync_plugin_license(
...
+			                        ) . '<br> ' . $this->get_text_inline( 'Error received from the server:', 'server-error-message' ) . var_export( $result->error, true ),
+			                        $this->get_text_x_inline( 'Oops', 'exclamation', 'oops' ) . '...',
+			                        'error',
+			                        $background,
+			                        'api_blocked'
+		                        );
+	                        }
                         } else {
+                            // API is working now. Delete the transient and start afresh.
+                            delete_transient('_fs_api_connection_retry_counter');
```
vovafeldman, line 21030: @msladu this part of the logic is executed when connectivity works but authentication params are wrong (which is rare and for edge cases). Please revise the method to position the transient deletion in a better place.

includes/class-freemius.php
vovafeldman: @msladu you didn't add clarification about why the counter is reset after 3 attempts (the first comment) in this thread.

includes/class-freemius.php
msladu: @vovafeldman Ok. Removing it.

includes/class-freemius.php
msladu: @vovafeldman Checking it.

includes/class-freemius.php
msladu: @vovafeldman Yes, my mistake. Changing the order.

includes/class-freemius.php
msladu: @vovafeldman Ok. I am adding that now.

includes/class-freemius.php
msladu: @vovafeldman Looking into it.

PR #532, opened 2021-12-28
Feature/hide billing and payments section

msladu:
- [account] Added filter `hide_billing_and_payments_info` to hide billing and payments section from account page.

### DIFF
--- a/templates/account/payments.php
+++ b/templates/account/payments.php
@@ -10,11 +10,15 @@
 		exit;
 	}
 
-	/**
-	 * @var array $VARS
-	 * @var Freemius $fs
-	 */
-    $fs       = freemius( $VARS['id'] );
+    /**
+     * @var array $VARS
+     * @var Freemius $fs
+     */
+    $fs = freemius( $VARS['id'] );
+
+    /**
+     * @var FS_Payment[] $payments
+     */
     $payments = freemius( $VARS['payments'] );
 
 	$slug = $fs->get_slug();
--- a/templates/account/payments.php
+++ b/templates/account/payments.php
@@ -14,15 +14,11 @@
 	 * @var array $VARS
 	 * @var Freemius $fs
 	 */
-	$fs = freemius( $VARS['id'] );
+    $fs       = freemius( $VARS['id'] );
+    $payments = freemius( $VARS['payments'] );
 
 	$slug = $fs->get_slug();
 
-	$payments = $fs->_fetch_payments();
-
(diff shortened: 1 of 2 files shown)


templates/account/payments.php
```diff
@@ -14,15 +14,11 @@
...
 	 */
-	$fs = freemius( $VARS['id'] );
+    $fs       = freemius( $VARS['id'] );
+    $payments = freemius( $VARS['payments'] );
 
 	$slug = $fs->get_slug();
 
-	$payments = $fs->_fetch_payments();
-
-	$show_payments = ( is_array( $payments ) && 0 < count( $payments ) );
```
msladu, line 23: @vovafeldman I have removed this code from here as we already fetching the data on accounts page. 

So, passing the data from accounts page here and retrieving the data using `freemius( $VARS['payments'] );`

vovafeldman: @msladu good catch 👍

templates/account/payments.php
```diff
@@ -14,15 +14,11 @@
 	 * @var array $VARS
 	 * @var Freemius $fs
 	 */
-	$fs = freemius( $VARS['id'] );
+    $fs       = freemius( $VARS['id'] );
+    $payments = freemius( $VARS['payments'] );
```
vovafeldman, line 18: @msladu for IDE autocompletion and improved documentation, unlike previously using `_fetch_payments()` the code could resolve the type of the variable because of the `@return` param of the method, now we need to explicitly specify it. We do it by adding a little comment above the relevant variable using the `@var` directive as following:
```
/**
 * @var FS_Payment[] $payments
 */
```

msladu: @vovafeldman Ok.

templates/account/payments.php
```diff
@@ -10,19 +10,19 @@
...
+     * @var array $VARS
+     * @var Freemius $fs
+     */
+    $fs = freemius( $VARS['id'] );
 
-	$slug = $fs->get_slug();
+    /**
+     * @var FS_Payment[] $payments
+     */
+    $payments = freemius( $VARS['payments'] );
```
vovafeldman, line 22: @msladu I just noticed you wrapped `$VARS['payments']` within `freemius()`, this is a mistake, it should simply be equal to `$VARS['payments']`.

msladu: @vovafeldman Will fix it.

PR #537, opened 2022-01-24
[php8.1] Silence the deprecation notices introduced in PHP 8.1

matczar:
There are several "PHP Deprecated" warnings in the `FS_Key_Value_Storage` class due to incompatibility with the `ArrayAccess` interface. These warnings appeared as of PHP 8.1: https://make.wordpress.org/core/2022/01/10/wordpress-5-9-and-php-8-0-8-1/#php-8-1-return-type-enforcement-and-new-returntypewillchange-attribute

### DIFF
--- a/includes/managers/class-fs-key-value-storage.php
+++ b/includes/managers/class-fs-key-value-storage.php
@@ -298,6 +298,7 @@
 			return $this->get( $k, null );
 		}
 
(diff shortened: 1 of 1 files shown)


vovafeldman:
Thank you @matczar 🙌

PR #548, opened 2022-03-09
Send Affiliate requests to the Bundle if the developer has used a bundle_id in dynamic_init

swashata:
Right now, we are sending affiliate requests to the main plugin installed through the bundle. Under the assumption that the developer would like to promote a Bundle when the `bundle_id` is set, we _should_ send the affiliate request to the bundle instead. This PR does that.

Additionally, I figured we could improve the application flow a little as we are already in the context.

---

## Improve error message shown to non-opted users

When a plugin/theme user is applying to become an affiliate, if they haven't opted in, they we do opt them, but we do not track.

However, the opt-in process does require consent from the email sent. Due to this, when a non-opted user tries to submit the affiliate form, it fails and just returns a message saying "Account is pending activation".

From a user's perspective who hasn't opted in (skipped during plugin activation), that does not give any instruction as to what to do next. So I improved the error message to give clear steps one needs to perform.

On the first step, it shows

> Account is pending activation. Please check your email and click the link to activate your account and then submit the affiliate form again.

Screen Shot 2022-03-09 at 1 07 01 PM

On the second step, the user goes to the email and clicks to opt-in (giving consent).

Screen Shot 2022-03-09 at 1 07 17 PM

On the third step, (after clicking the link), it takes to the Account page, stating that the account was activated.

Screen Shot 2022-03-09 at 1 07 33 PM

Finally, on the last step, the user submits the affiliate form again and a proper message is shown.

Screen Shot 2022-03-09 at 1 08 13 PM

---

## Fix the link to our User Dashboard

Instead of linking to `members.freemius.com` we link to `users.freemius.com`. I left the trailing slash `/` out from the URL intentionally because that is how the website works.

Screen Shot 2022-03-09 at 1 12 44 PM

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -10510,6 +10510,28 @@
         }
 
         /**
+         * Get whether the SDK has been initiated in the context of a Bundle.
+         *
+         * This will return true, if `bundle_id` is present in the SDK init parameters.
+         *
+         * ```php
+         * $my_fs = fs_dynamic_init( array(
+         *     // ...
+         *     'bundle_id'         => 'XXXX', // Will return true since we have bundle id.
+         *     'bundle_public_key' => 'pk_XXXX',
+         * ) );
+         * ```
+         *
+         * @author Swashata Ghosh (@swashata)
+         * @since  2.5.0
+         *
+         * @return bool True if we are running in bundle context, false otherwise.
+         */
+        private function has_bundle_context() {
+            return ! is_null( $this->get_bundle_id() );
+        }
+
+        /**
          * @author Vova Feldman (@svovaf)
          * @since  1.2.1.5
          *
@@ -14832,12 +14854,9 @@
          * @return number Bundle ID if developer has provided one, else the main plugin ID.
          */
         private function get_plugin_id_for_affiliate_terms() {
-            $bundle_id = $this->get_bundle_id();
-            if ( ! is_null( $bundle_id ) ) {
-                return $bundle_id;
-            }
-
-            return $this->_plugin->id;
+            return $this->has_bundle_context() ?
+                $this->get_bundle_id() :
+                $this->_plugin_id;
         }
 
         /**
@@ -14849,11 +14868,9 @@
                 /**
                  * In case we have a bundle set in SDK configuration, we would like to use that for affiliates, not the main plugin.
                  */
-                if ( ! is_null( $this->get_bundle_id() ) ) {
-                    $plugins_api = $this->get_api_bundle_scope();
-                } else {
-                    $plugins_api = $this->get_api_plugin_scope();
-                }
+                $plugins_api = $this->has_bundle_context() ?
+                    $this->get_api_bundle_scope() :
+                    $this->get_api_plugin_scope();
 
                 $affiliate_terms = $plugins_api->get( '/aff.json?type=affiliation', false );
 
@@ -23280,16 +23297,17 @@
 
             fs_enqueue_local_style( 'fs_affiliation', '/admin/affiliation.css' );
 
-            $is_bundle_context = ! is_null( $this->get_bundle_id() );
+            $is_bundle_context = $this->has_bundle_context();
 
             $plugin_title = $this->get_plugin_title();
+
             if ( $is_bundle_context ) {
                 $plugin_title = $this->plugin_affiliate_terms->plugin_title;
                 // Add the suffix "Bundle" only if the word is not present in the title itself.
                 if ( false === mb_stripos( $plugin_title, fs_text_inline( 'Bundle', 'bundle' ) ) ) {
                     $plugin_title = $this->apply_filters(
                         'formatted_bundle_title',
-                        $plugin_title . fs_text_inline( ' Bundle', 'bundle-suffix' )
+                        $plugin_title . ' ' . fs_text_inline( 'Bundle', 'bundle' )
                     );
                 }
             }
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -14825,14 +14825,41 @@
         }
 
         /**
+         * Get Plugin ID under which we will track affiliate application.
+         *
+         * This could either be the Bundle ID or the main plugin ID.
+         *
+         * @return number Bundle ID if developer has provided one, else the main plugin ID.
+         */
+        private function get_plugin_id_for_affiliate_terms() {
+            $bundle_id = $this->get_bundle_id();
+            if ( ! is_null( $bundle_id ) ) {
+                return $bundle_id;
+            }
+
+            return $this->_plugin->id;
+        }
+
+        /**
          * @author Leo Fajardo (@leorw)
          * @since  1.2.4
          */
         private function fetch_affiliate_terms() {
             if ( ! is_object( $this->plugin_affiliate_terms ) ) {
-                $plugins_api     = $this->get_api_plugin_scope();
+                /**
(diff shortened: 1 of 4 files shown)


includes/class-freemius.php
```diff
@@ -14823,15 +14823,42 @@ function has_affiliate_program() {
...
+        /**
+         * Get Plugin ID under which we will track affiliate application.
+         *
+         * This could either be the Bundle ID or the main plugin ID.
+         *
+         * @return number Bundle ID if developer has provided one, else the main plugin ID.
+         */
+        private function get_plugin_id_for_affiliate_terms() {
+            $bundle_id = $this->get_bundle_id();
+            if ( ! is_null( $bundle_id ) ) {
```
vovafeldman, line 14835: @swashata for short `if/else` returns I like using `?:`:
```
    return is_null( $bundle_id ) ?
        $this->_plugin->id :
        $bundle_id;
```

I think it improves readability.

includes/class-freemius.php
```diff
@@ -14823,15 +14823,42 @@ function has_affiliate_program() {
...
          * @author Leo Fajardo (@leorw)
          * @since  1.2.4
          */
         private function fetch_affiliate_terms() {
             if ( ! is_object( $this->plugin_affiliate_terms ) ) {
-                $plugins_api     = $this->get_api_plugin_scope();
+                /**
+                 * In case we have a bundle set in SDK configuration, we would like to use that for affiliates, not the main plugin.
+                 */
+                if ( ! is_null( $this->get_bundle_id() ) ) {
```
vovafeldman, line 14851: @swashata improvement suggestion to introduce a helper method named `has_bundle_context()` (or something along those lines) and use that instead of checking for `get_bundle_id()` every time.

vovafeldman: @swashata and I'd also recommend using `? :` instead of `if/else` here.

includes/class-freemius.php
```diff
@@ -14950,15 +14979,17 @@ function _submit_affiliate_application() {
                             var_export( $next_page, true )
                     );
                 } else if ( $this->is_pending_activation() ) {
-                    self::shoot_ajax_failure( $this->get_text_inline( 'Account is pending activation.', 'account-is-pending-activation' ) );
+                    self::shoot_ajax_failure( $this->get_text_inline( 'Account is pending activation. Please check your email and click the link to activate your account and then submit the affiliate form again.', 'account-is-pending-activation' ) );
```
vovafeldman, line 14982: @swashata good one!

includes/class-freemius.php
```diff
@@ -23248,7 +23279,24 @@ function _affiliation_page_render() {
 
             fs_enqueue_local_style( 'fs_affiliation', '/admin/affiliation.css' );
 
-            $vars = array( 'id' => $this->_module_id );
+            $is_bundle_context = ! is_null( $this->get_bundle_id() );
+
+            $plugin_title = $this->get_plugin_title();
+            if ( $is_bundle_context ) {
+                $plugin_title = $this->plugin_affiliate_terms->plugin_title;
+                // Add the suffix "Bundle" only if the word is not present in the title itself.
```
vovafeldman, line 23287: @swashata let's add a line space above to improve readability.

includes/class-freemius.php
```diff
@@ -23248,7 +23279,24 @@ function _affiliation_page_render() {
...
             fs_enqueue_local_style( 'fs_affiliation', '/admin/affiliation.css' );
 
-            $vars = array( 'id' => $this->_module_id );
+            $is_bundle_context = ! is_null( $this->get_bundle_id() );
+
+            $plugin_title = $this->get_plugin_title();
+            if ( $is_bundle_context ) {
+                $plugin_title = $this->plugin_affiliate_terms->plugin_title;
+                // Add the suffix "Bundle" only if the word is not present in the title itself.
+                if ( false === mb_stripos( $plugin_title, fs_text_inline( 'Bundle', 'bundle' ) ) ) {
```
vovafeldman, line 23288: @swashata can we assume `mb_` functions are available for all PHP setups?

includes/class-freemius.php
```diff
@@ -23248,7 +23279,24 @@ function _affiliation_page_render() {
...
+                    $plugin_title = $this->apply_filters(
+                        'formatted_bundle_title',
+                        $plugin_title . fs_text_inline( ' Bundle', 'bundle-suffix' )
```
vovafeldman, line 23291: @swashata I'd propose to use `fs_text_inline( 'Bundle', 'bundle' )` and add the space outside the transaction function. I understand how this provides more flexibility, but IMO it's too much. If someone will asks us for it, we can always enrich it later.

includes/i18n.php
```diff
@@ -143,7 +143,7 @@
         'paypal-email-address-is-required'   => _fs_text( 'PayPal email address is required.' ),
         'processing'                         => _fs_text( 'Processing...' ),
         'non-expiring'                       => _fs_text( 'Non-expiring' ),
-        'account-is-pending-activation'      => _fs_text( 'Account is pending activation.' ),
+        'account-is-pending-activation'      => _fs_text( 'Account is pending activation. Please check your email and click the link to activate your account and then submit the affiliate form again.' ),
```
vovafeldman, line 146: @swashata you can revert all the changes in the `i18n.php` file as it is no longer in use. Our previous approach was to have all the translations in the same file within the same array. But it was taking unnecessary RAM so we transitioned to "inline" translations. We kept it for backward compatibility. Now that it's been years already, it's safe to get rid of that file and usage. I'll create a task for it in Asana.

vovafeldman (changes requested):

includes/class-freemius.php
swashata: @vovafeldman The `server-environment` [handbook says](https://make.wordpress.org/hosting/handbook/server-environment/#php-extensions) that WordPress "highly recommends" `mbstring` extension. I did a little digging and it looks like [WordPress core has direct use of](https://core.trac.wordpress.org/browser/tags/5.9/src/wp-includes/functions.php#L5909) `mb_substr` with which we can assume that `mb_` functions are available for PHP setups where WordPress works. 

I used `mb_stripos` here (as you might've guessed already) to properly get the position of translated `"Bundle"` text from the plugin title. How about we keep this for now and come back at it only if it creates trouble?

includes/class-freemius.php
swashata: @vovafeldman I didn't do the helper method because the abstraction seemed a little hasty to me.

- It is called only at 2 places at the moment.
- In one place we do need the `bundle_id` and in the other place, we don't. 
 
So it'd become

```php
private function has_bundle_context() {
  return ! is_null( $this->get_bundle_id() );
}

private function get_plugin_id_for_affiliate_terms() {
  return $this->has_bundle_context() ?
    $this->get_bundle_id() : // also called from `has_bundle_context`
    $this->_plugin->id;
}

private function fetch_affiliate_terms() {
  // ...
  $plugins_api = $this->has_bundle_context() ?
    $this->get_api_bundle_scope() :
    $this->get_api_plugin_scope();
  // ...
}
```

But NP, I will make the changes as I can see how this could improve readability.

swashata: @vovafeldman Made the changes 👍

includes/class-freemius.php
vovafeldman: @swashata if core relies on it we can rely on it too.

vovafeldman (approved):
Looking good 👍

PR #550, opened 2022-03-14
Feature/anonymous license activation

msladu:
[permissions] Added diagnostic tracking allowed permission.


includes/class-freemius.php
```diff
@@ -17770,7 +17834,8 @@ private function install_with_new_user(
             $install_public_key,
             $install_secret_key,
             $redirect = true,
-            $auto_install = false
+            $auto_install = false,
+            $is_diagnostic_tracking_allowed = null
```
vovafeldman, line 17838: @msladu I suggest moving the new param to appear right after `$is_extensions_tracking_allowed` to logically appear next to the other customer permission (since it's a private method there's no need to think about backward compatibility here). Obviously, you also need to remember to update the calls.

This comment is also applicable to `install_many_pending_with_user()` and `install_many_with_new_user()`.

templates/connect.php
```diff
@@ -422,17 +434,15 @@ class="button button-secondary" tabindex="2"><?php fs_esc_html_echo_x_inline( 'S
...
-					<?php if ( $require_license_key ) : ?>
-						<p class="fs-license-sync-disclaimer"><?php
-                                echo sprintf(
-									fs_esc_html_inline( 'The %1$s will periodically send %2$s to %3$s for security & feature updates delivery, and license management.', 'license-sync-disclaimer', $slug ),
-									$fs->get_module_label( true ),
-									sprintf('<a class="fs-trigger" href="#" tabindex="1">%s</a>', fs_esc_html_inline('diagnostic data', 'send-data')),
-									'<a class="fs-tooltip-trigger' . (is_rtl() ? ' rtl' : '') . '" href="' . $freemius_site_url . '" target="_blank" rel="noopener" tabindex="1">freemius.com <i class="dashicons dashicons-editor-help" style="text-decoration: none;"><span class="fs-tooltip" style="width: 170px">' . $fs->get_text_inline( 'Freemius is our licensing and software updates engine', 'permissions-extensions_desc' ) . '</span></i></a>'
-								) ?></p>
-					<?php else : ?>
+                    <p class="fs-license-sync-disclaimer"><?php
```
vovafeldman, line 437: @msladu why did you remove the `if`?

vovafeldman (changes requested):

includes/class-freemius.php
```diff
@@ -5788,6 +5788,19 @@ function is_extensions_tracking_allowed() {
...
+        /**
+         * @author Malay Ladu (@msladu)
+         * @since  2.5.0.2
+         *
+         * @return bool
+         */
+        function is_diagnostic_tracking_allowed() {
+            return ( true === $this->apply_filters(
+                    'is_diagnostic_tracking_allowed',
+                    $this->_storage->get( 'is_diagnostic_tracking_allowed', null )
```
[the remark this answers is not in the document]
msladu: @vovafeldman Ok. Got it. Should I use the same for the  `is_extensions_tracking_allowed`?

includes/class-freemius.php
```diff
@@ -5783,23 +5783,23 @@ private function reconnect_locally( $is_context_single_site = false ) {
...
         }
 
-        /**
-         * @author Malay Ladu (@msladu)
-         * @since  2.5.0.2
-         *
-         * @return bool
-         */
-        function is_diagnostic_tracking_allowed() {
+	    /**
```
vovafeldman, line 5791: @malayladu please replace the tabs with spaces.

templates/connect.php
vovafeldman: @malayladu when activating a license key the permissions are triggered from a different link:
image

Please revert this change.

vovafeldman (changes requested):
@malayladu 
1. Please take a look at [this comment](https://github.com/Freemius/wordpress-sdk/pull/550#discussion_r836424471)
2. Please replace tabs with spaces (and make sure your IDE is configured to use spaces by default)
3. In addition, let's remove `'user_nickname'` in `Freemius::get_opt_in_params()`, as we don't have a property for storing the nickname on our end (and we don't need it).

swashata:
Before doing a code review, I am concerned about one behavior.

Let us consider the scenario, I want to activate without changing anything.

## What happens in the FREE version activation scenario (both develop and this branch):

(1) SDK Version and Plugin Version is always turned on. There is no way to deactivate it. I understand this is the intended behavior.

(2) There is no section for "Title, WP Version, Locale & PHP Version", i.e diagnostic information.

(3) PLUGINS & THEMES are **TURNED ON** by default.

image

## What happens in the LICENSED version activation scenario:

(1) SDK Version and Plugin Version is always turned on.

(2) There is a section for the Diagnostic Information, but it is "TURNED OFF** by default.

(3) PLUGINS & THEMES are **TURNED OFF** by default.

image

---

If not explicitly intended, I believe the behavior in the LICENSE activation scenario is already buggy and we are adding more debt to it.

1. It keeps "Plugins & Theme" turned off by default, not consistent with the free version activation.
2. It now introduces a new "Diagnostic Info" section, but doesn't introduce it for the free version too. In the free version activation, it assumes those are turned on.
3. It actually DOES NOT send the "Plugins & Themes" and the now introduced "Diagnostic Info" to our API, as those were turned off.

So I propose the following workflow.

1. Free version or not, show the two "Diagnostic Info" & "Plugins & Themes" with toggles "turned on" by default.
2. Depending on "Free version" or "License" activation workflow, show those sections when trigger is clicked.
3. If user makes changes to the toggles, then obey it.

Please let me know your thought.

fajardoleo:
@vovafeldman In `/templates/js/permissions.php` file, we have this method:
```function setUpdatingLabel( $button, isEnabled ) {
    $button.text( isEnabled ?
        '<?php fs_esc_js_echo_inline( 'Opting in', 'opting-in' ) ?>...' :
        '<?php fs_esc_js_echo_inline( 'Opting out', 'opting-out' ) ?>...'
    );
}
``` 

When `isEnabled` is `false`, the label on the Opt Out dialog is `Opt In`, so when clicking on it, the label becomes `Opting out...` instead of `Opting in...`.

fajardoleo (changes requested):

vovafeldman (approved):

fajardoleo (changes requested):

PR #554, opened 2022-03-21
[translation] [deprecation] [code removal] Removed deprecated i18n fi…

drory82:
[translation] [deprecation] [code removal] Removed deprecated i18n file, global hard-coded translation array and its deprecated usages


includes/fs-essential-functions.php
```diff
@@ -149,84 +149,6 @@ function fs_kses_no_null( $string ) {
 
 	#endregion Core Redirect (copied from BuddyPress) -----------------------------------------
 
-	if ( ! function_exists( '__fs' ) ) {
-		global $fs_text_overrides;
-
-		if ( ! isset( $fs_text_overrides ) ) {
```
vovafeldman, line 155: @drory82 from inspecting the code, this is the only place where `$fs_text_overrides` was initialized. So, when calling `fs_text()` (or any of the other methods that rely on that function), it will throw a PHP warning when executing the following if (file: `includes/fs-core-functions.php`):
`if ( isset( $fs_text_overrides[ $slug ] ) ) {`

So my suggestion is to move this initialization to `includes/fs-core-functions.php` as the 1st line within `if ( ! function_exists( 'fs_text' ) ) {`.

P.S. You also need to move the `global $fs_text_overrides;` so the script will know you are initializing the global var and not a local scope one.

vovafeldman (changes requested):
In addition to the inline comment, please also update `gulpfile.js`: 
* There's a reference to the deprecated `i18n.php` file in one of the comments - let's update it to `Create POT out of PHP files.` (that's what the gulp task does there).
* Let's get rid of the commented code (related to the legacy system):
  * https://github.com/Freemius/wordpress-sdk/blob/master/gulpfile.js#L34-L39
  * https://github.com/Freemius/wordpress-sdk/blob/master/gulpfile.js#L59-L76

includes/fs-essential-functions.php
drory82: This function uses it as well: fs_text_override
So as far as I understand, it won't be enough to init it in fs_text, in case fs_text_override will be called before it.
Am I correct ?

vovafeldman: @drory82 yes, you got it right 👍

drory82: What do you suggest, doing something like this before the declaration of any of the 3 methods:
image

Or, is there a global place I better declare and init it ?

drory82: BTW, I think it would be nicer to put it before the documentation of the function it self.

drory82: @vovafeldman, Thanks. Done

vovafeldman: @drory82 I think there was some confusion - I meant to only add it once before the declaration of `fs_text()`. 

Having said that, after thinking it through again, it's probably safer to place before the `if ( ! function_exists( 'fs_text' ) ) {` if we want to include it once.

vovafeldman (approved):
Looks good now 👍

includes/fs-essential-functions.php
drory82: @vovafeldman  Ye, this is why I've asked if "is there a global place where I can declare and init once..."
But, it is still confusing me that all the code is just running, even between method declaration :) I'm used to function declaration only within a class I guess...
Anyhow, I hope it is ok now.

PR #555, opened 2022-03-29
[connect] Include the "skip" part of the message only if anonymous mo…

fajardoleo:
…de is supported.


vovafeldman (approved):
Looks good 👍

PR #559, opened 2022-04-19
[pricing] Added an option to change the bulk-site discounts model on …

fajardoleo:
…the pricing page.

### DIFF
--- a/templates/pricing.php
+++ b/templates/pricing.php
@@ -88,7 +88,8 @@
 		// Billing cycle.
 		'billing_cycle'    => fs_request_get( 'billing_cycle', WP_FS__PERIOD_ANNUALLY ),
 		'is_network_admin' => fs_is_network_admin() ? 'true' : 'false',
-		'currency'         => $fs->apply_filters( 'default_currency', 'usd' ),
+        'currency'         => $fs->apply_filters( 'default_currency', 'usd' ),
+        'discounts_model'  => $fs->apply_filters( 'pricing/discounts_model', 'absolute' ),
 	) );
 
     $use_external_pricing = $fs->should_use_external_pricing();
@@ -123,7 +124,6 @@
         <?php
         $pricing_config = array_merge( array(
             'contact_url'            => $fs->contact_url(),
-            'is_network_admin'       => fs_is_network_admin(),
             'is_production'          => ( defined( 'WP_FS__IS_PRODUCTION_MODE' ) ? WP_FS__IS_PRODUCTION_MODE : null ),
             'menu_slug'              => $fs->get_menu_slug(),
             'mode'                   => 'dashboard',
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.5.0.6';
+	$this_sdk_version = '2.5.0.7';
 
 	#region SDK Selection Logic --------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


templates/pricing.php
```diff
@@ -88,7 +88,8 @@
 		// Billing cycle.
 		'billing_cycle'    => fs_request_get( 'billing_cycle', WP_FS__PERIOD_ANNUALLY ),
 		'is_network_admin' => fs_is_network_admin() ? 'true' : 'false',
-		'currency'         => $fs->apply_filters( 'default_currency', 'usd' ),
+        'currency'         => $fs->apply_filters( 'default_currency', 'usd' ),
+        'discounts_model'  => $fs->apply_filters( 'pricing/discounts_model', 'absolute' ),
```
swashata, line 92: @fajardoleo , I am thinking about two things here.

## 1: Introduce a `FILTERS.md` documentation

I know we already have lots of filters, which are kind of undocumented. But let us try fix that with this PR.

- Create a FILTERS.md file in the root of the repository.
- Please mention that this is a WIP and we have started documenting new filters after (28th Apr 2022, or the date you manage to create the file). Also kindly mention that we are planning to add existing filters as we encounter and contributions are welcome.
- Do add the usage of this filter with examples and also kindly mention how it would work with the React Pricing App and the iframe pricing page.

## 2: Structure the acceptable values of the `pricing/discounts_model` filter

Ideally I would have preferred if we had a bunch of "contract" or "preset" classes. Like:

```php
// somewhere in `includes/Contracts/Filters/Pricing`
namespace FreemiusWPSDK\Contracts\Filters\Pricing;

class DiscountsModel {
    const RELATIVE = 'relative';
    const ABSOLUTE = 'absolute';

    const FILTER = 'pricing/discounts_model';
}

// in `templates/pricing.php`
   'discounts_model' => $fs->apply_filters( DiscountsModel::FILTER, DiscountsModel::ABSOLUTE ),
```

Just think how easy it'd become then, to document and use the filters without having to remember the names and values everywhere. This would also expose a better "SDK" usage for our users.

That being said, I understand it could be too much structural change to aim for at the moment. We know WP now supports only PHP 5.6 at the moment, so we would be good on the namespace front. But we might need to heavily modify our "auto-loader" and "detect latest freemius sdk" functionalities.

So for the time being, let us structure the accepted value of the `pricing/discounts_model`, through some constants.

```php
// in the same file where other constants are defined.
define( 'WP_FS_FILTER_PRICING_DISCOUNT_MODEL_ABSOLUTE', 'absolute' );
define( 'WP_FS_FILTER_PRICING_DISCOUNT_MODEL_RELATIVE', 'relative' );

// in templates/pricing.php
'discounts_model'  => $fs->apply_filters( 'pricing/discounts_model', WP_FS_FILTER_PRICING_DISCOUNT_MODEL_ABSOLUTE )
```

Then also in the `FILTERS.md`

```md
The accepted values for the filter `pricing/discounts_model` are (pre-defined PHP constants available through our SDK:

- `WP_FS_FILTER_PRICING_DISCOUNT_MODEL_ABSOLUTE` - Will show absolute discounts in the ....
- `WP_FS_FILTER_PRICING_DISCOUNT_MODEL_RELATIVE` - Will show relative discounts ...
```

swashata (commented):
@fajardoleo Please see my comments on this Merged PR and kindly open another PR with the proposed fixes.

PR #560, opened 2022-04-27
Feature/license recovery message

xiahengchen:
- [license-recovery] [message] [update] Updated the license recovery message to show slightly different when the context product is paid only.

### DIFF
--- a/templates/forms/resend-key.php
+++ b/templates/forms/resend-key.php
@@ -54,7 +54,10 @@ HTML;
 HTML;
 	}
 
-	$message_above_input_field = fs_esc_html_inline( "Enter the email address you've used " . $fs->is_only_premium() ? 'during the purchase' : 'for the upgrade below' . ' and we will resend you the license key.', 'ask-for-upgrade-email-address', $slug );
+	$message_above_input_field = $fs->is_only_premium() ?
+        fs_esc_html_inline( "Enter the email address you've used during the purchase and we will resend you the license key.", 'ask-for-upgrade-email-address-premium-only', $slug ) :
+        fs_esc_html_inline( "Enter the email address you've used for the upgrade below and we will resend you the license key.", 'ask-for-upgrade-email-address', $slug );
+
 	$modal_content_html = <<< HTML
     <div class="notice notice-error inline license-resend-message"><p></p></div>
     <p>{$message_above_input_field}</p>
--- a/templates/forms/resend-key.php
+++ b/templates/forms/resend-key.php
@@ -54,7 +54,7 @@ HTML;
 HTML;
 	}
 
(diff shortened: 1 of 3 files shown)


templates/forms/resend-key.php
```diff
@@ -54,7 +54,7 @@
 HTML;
 	}
 
-	$message_above_input_field = fs_esc_html_inline( "Enter the email address you've used for the upgrade below and we will resend you the license key.", 'ask-for-upgrade-email-address', $slug );
+	$message_above_input_field = fs_esc_html_inline( "Enter the email address you've used " . $fs->is_only_premium() ? 'during the purchase' : 'for the upgrade below' . ' and we will resend you the license key.', 'ask-for-upgrade-email-address', $slug );
```
swashata, line 57: @xiahengchen , I am not sure whether these has been done or not, but from now on, let us **NOT** use interpolated strings inside `fs_esc_html_inline`. Those strings are supposed to be used for translations and they need to be "non interpolated".

So instead of what you've done, we need to do something like this:

```php
$message_above_input_field = $fs->is_only_premium() ?
    fs_esc_html_inline( 'full text in single quote goes here', 'ask-for-upgrade-email-address-premium-only' ) :
    fs_esc_html_inline( '...', 'ask-for-upgrade-email-address' );

```

swashata (changes requested):
@xiahengchen Please make the change.

templates/forms/resend-key.php
```diff
@@ -54,7 +54,10 @@
 HTML;
 	}
 
-	$message_above_input_field = fs_esc_html_inline( "Enter the email address you've used for the upgrade below and we will resend you the license key.", 'ask-for-upgrade-email-address', $slug );
+	$message_above_input_field = $fs->is_only_premium() ?
+        fs_esc_html_inline( "Enter the email address you've used during the purchase and we will resend you the license key.", 'ask-for-upgrade-email-address-premium-only', $slug ) :
```
swashata, line 58: @xiahengchen, please add indentation here (4 spaces).

```php
$message_above_input_field = $fs->is_only_premium() ?
    fs_esc_html_inline( "..." ) :
    fs_esc_html_inline( "..." );
```

Also the file seems to have mixed indentation. Please change the indentation of the touched lines to 4 spaces (includes the tab before the variable `$message_above_input_field`.

swashata (changes requested):
@xiahengchen , please make the changes along with the followings and we can merge this:

1. Change the version number of [`start.php`](https://github.com/Freemius/wordpress-sdk/blob/develop/start.php#L18) from `2.5.0.7` to `2.5.0.8`.
2. While we are at it, the version of [`package.json`](https://github.com/Freemius/wordpress-sdk/blob/develop/package.json#L7) says `2.5.0.1`. This is an invalid semver that I mistakenly committed. Since this `package.json` is only used to install our own dev dependencies and is not used anywhere, let's change this to `0.0.0` and add `"private": true`.

swashata (approved):
LGTM 👏

PR #561, opened 2022-04-27
Feature/delete account

xiahengchen:
- [account] [delete] [messages] [update] Updated the term 'Delete Account' to 'Disconnect' in the relevant fields since it is the detachment of the account, not the account deletion.


templates/account.php
```diff
@@ -294,12 +294,12 @@ class="nav-tab"><?php echo esc_html( $addons_text ) ?></a>
                                                     <?php wp_nonce_field( 'delete_account' ) ?>
                                                     <a class="fs-delete-account" href="#" onclick="if (confirm('<?php
```
swashata, line 295: @xiahengchen 

Since we are touching this, let's fix some of the semantics here too. I would love for you to:

- [ ] Create a sub template for this `if ( ! $is_paying ) : ` under `templates/account/is-paying-disconnect-button.php` and move the code there.

**templates/account/is-paying-disconnect-button.php**
```php
<?php
/**
 * @var Freemius $fs
 */
?>
<form action="<?php echo esc_attr( $fs->_get_admin_page_url( 'account' ); ?>" method="POST">
 ....
</form>
```

Do mind the usage of `esc_attr` and [other escaping functions](https://developer.wordpress.org/plugins/security/securing-output/#escaping).

In the `if` block of `templates/account.php` simply `require` (not include) that sub template file.

- [ ] Change the anchor to a button. Do add relevant CSS for it to be in the same style.

```html
<?php
// On top of the template file, calculate in advance the ID of the button.
$button_id = 'fs-disconnect-button-' . $fs->get_id();
// Also the message.
$message = $is_active_subscription ?
     fs_text_inline( ... ) :
     fs_text_inline( ... );
?>
<button class="fs-delete-account fs-button-inline" id="<?php echo esc_attr( $button_id ); ?>"> ...</button>
```

- [ ] Add inline JavaScript in the sub template file for handling the click.

```html
<script type="text/javascript">
// Wrap in a IFFE to prevent leaking global variables.
(function() {
  var disconnect_button_id = <?php echo wp_json_encode( $button_id ); ?>;
  var message = <?php echo wp_json_encode( $message ); ?>;

  jQuery(document).ready(function($) {
      $('#' + disconnect_button_id).click(function(e) {
          // Prevent the form being submitted.
          e.preventDefault();
          // Show the confirm and if yes, then submit the form.
      });
  });
})();
</script>
```

This will probably take you 30mins to 1hr to implement, but we will have a better codebase 😃 . If we follow this strategy for all PRs, then eventually our codebase will become better 👍

swashata (changes requested):
@xiahengchen Please make the changes.

assets/scss/admin/account.scss
```diff
@@ -73,6 +73,29 @@
             }
         }
     }
+
+    .fs-button-inline
+    {
+        background-color: white;
```
swashata, line 79: - [ ] Let' make it `background-color: transparent`. 

- [ ] And our convention to write SCSS is

```scss
.fs-button-inline {
    // ...
}
```

- [ ] Let's move this styling to a separate file for re-usability. [This file](https://github.com/Freemius/wordpress-sdk/blob/master/assets/scss/admin/_buttons.scss) seems like the correct place.

templates/account.php
```diff
@@ -289,18 +289,14 @@ class="nav-tab"><?php echo esc_html( $addons_text ) ?></a>
...
-                                                            class="dashicons dashicons-no"></i> <?php fs_esc_html_echo_inline( 'Delete Account', 'delete-account', $slug ) ?></a>
-                                                </form>
+                                                <?php
+                                                    $view_params = array(
+                                                        'freemius'               => $fs,
+                                                        'is_active_subscription' => $is_active_subscription,
+                                                        'slug'                   => $slug,
+                                                        'plan'                   => $plan
+                                                    );
+                                                    fs_require_template( 'account/is-paying-disconnect-button.php', $view_params ); ?>
```
swashata, line 299: Awesome 👏

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
swashata, line 1: @xiahengchen Very good 👏

swashata (changes requested):
@xiahengchen, excellent job. Please finalize the small changes and it would be ready.

templates/account.php
```diff
@@ -289,18 +289,14 @@ class="nav-tab"><?php echo esc_html( $addons_text ) ?></a>
...
-                                                        } else {
-                                                            echo esc_attr( sprintf( fs_text_inline( 'Deletion is not temporary. Only delete if you no longer want to use this %s anymore. Are you sure you would like to continue with the deletion?', 'delete-account-confirm', $slug ), $fs->get_module_label( true ) ) );
-                                                        }
-                                                    ?>'))  this.parentNode.submit(); return false;"><i
-                                                            class="dashicons dashicons-no"></i> <?php fs_esc_html_echo_inline( 'Delete Account', 'delete-account', $slug ) ?></a>
-                                                </form>
+                                                <?php
+                                                    $view_params = array(
+                                                        'freemius'               => $fs,
+                                                        'is_active_subscription' => $is_active_subscription,
```
swashata, line 295: @xiahengchen I started testing this out. Turns out the `$is_paying` and `$is_active_subscription` are mutually inclusive. When `$is_paying` is `false`, the `$is_active_subscription` will also be `false`. Our check works like that.

So when a license is active, i.e, `$is_active_subscription`, we don't actually show this block. Instead we show this.

image

Click on the "Deactivate License" button actually tells us a similar message.

image

So, we don't need this extra check and this extra variable. It was an oversight from our end. Hence, kindly remove this variable from template params.

Instead, what we do want to introduce is, if there's a license available. Like check the state of account here. I can activate the "Ultra License" because it was deactivated, but not yet fully disconnected.

image

From the code, the check works like this:

```php
$available_license = $fs->is_free_plan() && ! fs_is_network_admin() ? $fs->_get_available_premium_license( $site->is_localhost() ) : false;
```

You will find this on line `545` of this file. So let us move the `$available_license` logic on the top and pass `is_license_available` to the template. If `$is_license_available` is present, then change the `$message`.

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,53 @@
...
+    if ( ! defined( 'ABSPATH' ) ) {
+        exit;
+    }
+
+    /**
+     * @var array $VARS
+     * @var Freemius $fs
+     */
+    $fs                     = $VARS['freemius'];
+    $is_active_subscription = $VARS['is_active_subscription'];
```
swashata, line 18: As mentioned in my previous review, we need to remove this, and introduce `$is_license_available` here and make the check against that.

swashata (changes requested):
@xiahengchen I found our check was really flawed here 😞 . I have added comments on why and what we need to do instead. It is still a small change. Please look into it.

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,55 @@
...
+    $fs                 = $VARS['freemius'];
+    $premium_plan_title = $VARS['premium_plan_title'];
+    $slug               = $VARS['slug'];
+
+    $button_id = "fs-disconnect-button-{$fs->get_id()}";
+
+    $product_type = $fs->get_module_label( true );
+
+    $message = $premium_plan_title ?
+        sprintf( fs_text_inline( 'Your %s plan is not active on this site at the moment and you can activate it on any other site. Only disconnect if you no longer want to use this %s anymore. If you wish to stop the recurring payment of the %s plan instead, then kindly click the CANCEL button, activate the %s plan and click on the "Downgrade" button to downgrade your account. Are you sure you would like to continue with this disconnection?', 'disconnect-x-confirm', $slug ), $premium_plan_title, $product_type, $premium_plan_title, $premium_plan_title ) :
```
swashata, line 26: @xiahengchen Please use positioned placeholder. With translators comments (also valid for the next line).

```php
/** translators: %1$s is replaced by the premium plan name, %2$s is replaced by the type of the module */
sprintf( fs_text_inline( 'Your %1$s plan is not active on this site at the moment and you can activate it on any other site. Only disconnect if you no longer want to use this %2$s anymore. If you wish to stop the recurring payment of the %1$s plan instead, then kindly click the CANCEL button, activate the %1$s plan and click on the "Downgrade" button to downgrade your account. Are you sure you would like to continue with this disconnection?', 'disconnect-x-confirm', $slug ), $premium_plan_title, $product_type )
```

swashata (changes requested):
@xiahengchen everything looks good. We just need to change the sprintf placeholder to avoid repeated arguments and add translators comments.

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+     */
+    $fs                 = $VARS['freemius'];
+    $premium_plan_title = $VARS['premium_plan_title'];
+    $slug               = $VARS['slug'];
+
+    $button_id = "fs-disconnect-button-{$fs->get_id()}";
+
+    $product_type = $fs->get_module_label( true );
+
+    /** translators: %1$s is replaced by the premium plan name, %2$s is replaced by the type of the module */
```
swashata, line 25: This translators comment needs to be placed before the sprintf.

swashata (changes requested):

templates/account.php
```diff
@@ -243,6 +243,10 @@
     $addons_to_show = array_unique( array_merge( $installed_addons_ids, $account_addons ) );
 
     $is_active_bundle_subscription = ( is_object( $bundle_subscription ) && $bundle_subscription->is_active() );
+
+    $available_license = $fs->is_free_plan() && ! fs_is_network_admin() ? $fs->_get_available_premium_license($site->is_localhost()) : false;
```
vovafeldman, line 247: @swashata let's change the formatting so it will be easier to digest:
```
$available_license = ( $fs->is_free_plan() && ! fs_is_network_admin() ) ?
    $fs->_get_available_premium_license( $site->is_localhost() ) :
    false;
```

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+    $message = $premium_plan_title ?
+        sprintf( fs_text_inline( 'Your %1$s plan is not active on this site at the moment and you can activate it on any other site. Only disconnect if you no longer want to use this %2$s anymore. If you wish to stop the recurring payment of the %1$s plan instead, then kindly click the CANCEL button, activate the %1$s plan and click on the "Downgrade" button to downgrade your account. Are you sure you would like to continue with this disconnection?', 'disconnect-x-confirm', $slug ), $premium_plan_title, $product_type ) :
```
vovafeldman, line 27: @swashata the copy of the prev messages is very old, most likely before we even have the User Dashboard. I think the instructions to cancel the subscription can be simplified. Here's what I propose to use instead:
```
Disconnecting the website will permanently remove it from your User Dashboard's account. If you wish to cancel your %1$s plan's subscription instead, please navigate to the User Dashboard (https://users.freemius.com) and cancel it there. Are you sure you would like to disconnect the website permanently?
```

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+        sprintf( fs_text_inline( 'Only disconnect if you no longer want to use this %s anymore. Are you sure you would like to continue with the disconnection?', 'disconnect-confirm', $slug ), $product_type );
+
+?>
+<script type="text/javascript">
+    // Wrap in a IFFE to prevent leaking global variables.
+    (function() {
+        var disconnect_button_id = <?php echo wp_json_encode( $button_id ); ?>;
```
vovafeldman, line 34: @swashata WP JS variables convention is `camelCase`. Also, when there's only a single line of PHP, we avoid adding `;` at the end of the PHP line.

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+                // Prevent the form being submitted.
+                e.preventDefault();
+
+                // Show the confirm and if yes, then submit the form.
+                if (confirm(message)) {
+                    this.parentNode.submit();
+                }
+            });
+        });
+    })();
```
vovafeldman, line 48: @swashata It is not required but a good practice to pass `jQuery` as an argument and then have a `$` param in the wrapping function like this:
```
(function($) {
    // Now you can use $ for jQuery access.
})(jQuery);
```

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+        sprintf( fs_text_inline( 'Only disconnect if you no longer want to use this %s anymore. Are you sure you would like to continue with the disconnection?', 'disconnect-confirm', $slug ), $product_type );
+
+?>
+<script type="text/javascript">
+    // Wrap in a IFFE to prevent leaking global variables.
+    (function() {
+        var disconnect_button_id = <?php echo wp_json_encode( $button_id ); ?>;
+        var message = <?php echo wp_json_encode( $message ); ?>;
+
+        jQuery(document).ready(function($) {
```
vovafeldman, line 37: @swashata WP JS formatting is add spacing after and before `'(', ')'`. Example:
```
$( '#' + disconnect_button_id ).click( function( e ) {
```

We only started to use that properly ~1yr ago, so you'll find old JS formatted differently. But I think it's good to follow WP practices going forward for the SDK.

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @var array $VARS
+     * @var Freemius $fs
+     */
+    $fs                 = $VARS['freemius'];
+    $premium_plan_title = $VARS['premium_plan_title'];
+    $slug               = $VARS['slug'];
+
+    $button_id = "fs-disconnect-button-{$fs->get_id()}";
```
vovafeldman, line 21: @swashata unlike `class` attributes, the convention for HTML element `id`s is separating words with an underscore.

vovafeldman (commented):
@swashata ready for your review.

templates/account/is-paying-disconnect-button.php
```diff
@@ -0,0 +1,56 @@
...
+                }
+            });
+        });
+    })();
+</script>
+<form action="<?php echo esc_attr( $fs->_get_admin_page_url( 'account' )); ?>" method="POST">
+    <input type="hidden" name="fs_action" value="delete_account">
+    <?php wp_nonce_field( 'delete_account' ) ?>
+
+    <button class="fs-button-inline" id="<?php echo esc_attr( $button_id ); ?>"><i
```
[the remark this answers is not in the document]
swashata: Because `a` is an anchor, that is used for navigation, while `button` is used for action. Please do let me know if we want to continue with HTML semantics, the added CSS is not very big and we can slowly start transitioning away from using `a` where `button` makes more sense.

templates/account/is-paying-disconnect-button.php
swashata: From my old days, I had a different practice 😄 

```js
jQuery(document).ready(function($) {
    // Now you can use $ for jQuery access.
});
```

But yes, IFFE parameters are equally valid and if this is the convention, I will remember it 👍

swashata:
@vovafeldman Please see my one unresolved comment and make a call. I have fixed the rest.

templates/account/is-paying-disconnect-button.php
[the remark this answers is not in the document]
vovafeldman: @swashata let's revert to `<a>`. While I get and agree with your point, as we are already getting complaints about the size of the SDK, I feel the size is more important than the semantics here.

PR #567, opened 2022-07-04
Feature/multisite bulk sites license activation fix

fajardoleo:


vovafeldman (changes requested):
Looks good, @fajardoleo. The only thing that I'd recommend to do is create a `$items_per_request = 25;` helper var and replace it with the 3 occurrences of the 25 in that function.

PR #572, opened 2022-07-22
[options] [cache] Improved the options caching logic.

fajardoleo:


includes/managers/class-fs-option-manager.php
```diff
@@ -437,11 +437,31 @@ function store() {
...
+            if ( WP_FS__DEBUG_SDK ) {
+                return false;
+            }
+
+            if ( defined( 'WPMUDEV_HOSTING_OBJECT_CACHE_ENABLED' ) && WPMUDEV_HOSTING_OBJECT_CACHE_ENABLED ) {
+                return false;
+            }
+
+            return (
+                ! defined( 'WP_FS__CACHE_ENABLED' ) || WP_FS__CACHE_ENABLED
```
vovafeldman, line 461: @fajardoleo, since the cache would be on by default, I propose to change the flag to `WP_FS__CACHE_DISABLED`. In addition, let's add support for another flag `WP_FS__OBJECT_CACHE_DISABLED` and when set to `true`, check if [wp_using_ext_object_cache()](https://developer.wordpress.org/reference/functions/wp_using_ext_object_cache/) is also `true` and return `false` (i.e., this flag will disable the cache only when object cache is enabled).

Additionally, let's add those flags with a default `false` value to the `config.php` file so people will be able to see all the flags the SDK supports.

vovafeldman (changes requested):

includes/managers/class-fs-option-manager.php
vovafeldman: I did some further thinking, and it would be good to enrich our Debug page with:
* Showing the state of the different cache values: `WPMUDEV_HOSTING_OBJECT_CACHE_ENABLED`, `WP_FS__CACHE_DISABLED`, `WP_FS__OBJECT_CACHE_DISABLED`, `wp_using_ext_object_cache()`
* Introduce two on/off switches to allow overriding the values of `WP_FS__CACHE_DISABLED`, `WP_FS__OBJECT_CACHE_DISABLED`.

This will allow us to debug future caching issues much easier, as we can simply ask users to send us a screenshot of the debug page. Plus, we can guide them very easily to turn off the cache by clicking a switch without the need to make any code changes.

vovafeldman:
This PR is on hold. The following commits is a candidate for a better solution to the caching issue:
* https://github.com/Freemius/wordpress-sdk/commit/c83f2df24cedb2f53baa14022e2fee000bd9f641
* https://github.com/Freemius/wordpress-sdk/commit/2dc303a9d4fec0a4633cda90ced765fad8857977

PR #588, opened 2022-11-01
Feature/clone resolution enhancements

fajardoleo:


includes/class-freemius.php
```diff
@@ -3461,9 +3467,32 @@ function is_clone() {
...
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.1
+         */
+        function maybe_update_clone_resolution_support_flag() {
+            if ( isset( $this->_storage->has_clone_resolution_support ) ) {
+                return;
+            }
+
+            $this->_storage->has_clone_resolution_support = (
```
vovafeldman, line 3493: @fajardoleo I think the name selected for the flag is misleading since the SDK does support clone resolution. I feel a better var name would be `hide_manual_resolution` (i. opposite meaning; ii. once this moved into the clone manager class it will use its storage, so the clone context is obvious and there's no need to include it in the name of the property), and instead of `$is_strict_mode` rename to `$only_if_manual_resolution_is_not_hidden`.

includes/class-freemius.php
```diff
@@ -9789,6 +9845,57 @@ private function get_install_diff_for_api( $site, $install, $override = array()
...
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since  2.5.1
+         *
+         * @param FS_Site  $site
+         * @param string[] $params
+         */
+        function maybe_send_clone_update( $site = null, $params = array(), $initial_update = false ) {
```
vovafeldman, line 9855: @fajardoleo
* Let's split it into 2 methods
* Let's remove the `$this->is_tracking_allowed()` check (the intent of the resolution is there when attempting to resolve regardless of the opt-in permissions)

vovafeldman (changes requested):

includes/entities/class-fs-site.php
```diff
@@ -234,4 +234,23 @@ function is_trial_utilized() {
...
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.1
+         *
+         * @param int $blog_id
+         *
+         * @return bool
+         */
+        function is_clone( $blog_id = null ) {
+            $blog_id = ( empty( $blog_id ) && is_multisite() && ! empty( $this->blog_id ) ) ?
```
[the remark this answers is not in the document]
fajardoleo: @vovafeldman Another call is in `/templates/debug.php` file (`$site->is_clone()`). Should we introduce a `$site_url` parameter instead?

includes/class-freemius.php
fajardoleo: @vovafeldman Is the `is_clone()` method supposed to be also moved to the clone manager class?

includes/entities/class-fs-site.php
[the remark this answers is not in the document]
vovafeldman: @fajardoleo yes! going with `$site_url` is a good idea and will also eliminate the dependency on the dynamic `blog_id` prop.

includes/class-freemius.php
vovafeldman: @fajardoleo I think it will complicate things since the logic relies on `$this->_site`, which means that we'll need to load the site every time before calling it. So my intuition is to keep it in `Freemius`.

includes/managers/class-fs-clone-manager.php
```diff
@@ -817,41 +847,54 @@ private function resolve_cloned_sites( $clone_action, $fs_instances = array() )
...
 
-            if ( self::OPTION_TEMPORARY_DUPLICATE === $clone_action ) {
-                $this->store_temporary_duplicate_timestamp();
-            } else {
-                $instances_with_clone_count = 0;
-                $instance_with_error        = null;
-                $has_error                  = false;
+            $instances_with_clone       = array();
+            $instances_with_clone_count = 0;
+            $instance_id_to_install_map = array();
```
vovafeldman, line 852: @fajardoleo I suggest renaming to `$install_by_instance_id`

templates/debug.php
```diff
@@ -112,6 +116,15 @@
...
             <button id="fs_set_db_option" class="button"><?php fs_esc_html_echo_inline( 'Set DB Option' ) ?></button>
         </td>
+        <td>
+            <?php
+                $fs_debug_page_url = 'admin.php?page=freemius&fs_action=allow_clone_resolution_notice';
+                $fs_debug_page_url = fs_is_network_admin() ?
+                    network_admin_url( $fs_debug_page_url ) :
+                    admin_url( $fs_debug_page_url );
+            ?>
+            <a href="<?php echo $fs_debug_page_url ?>" class="button button-primary">Resolve Clone(s)</a>
```
vovafeldman, line 126: @fajardoleo we need to add nonce protection here.

includes/class-freemius.php
```diff
@@ -17088,12 +17164,13 @@ private function _load_account() {
...
-                empty( FS_Clone_Manager::instance()->get_clone_identification_timestamp() )
-            ) {
-                FS_Clone_Manager::instance()->store_clone_identification_timestamp();
+            if ( $this->is_user_in_admin() && $this->is_clone() ) {
+                if ( empty( FS_Clone_Manager::instance()->get_clone_identification_timestamp() ) ) {
+                    FS_Clone_Manager::instance()->store_clone_identification_timestamp();
+                }
+
+                FS_Clone_Manager::instance()->maybe_update_clone_resolution_support_flag( $this->_storage->sdk_last_version );
+                $this->send_pending_clone_update();
```
vovafeldman, line 17173: @fajardoleo I'd propose to add `_once` to the end of the method's name making it clear it only happens once.

vovafeldman (changes requested):

vovafeldman (approved):

Issue #513, opened 2021-07-30
PHP Notice: Undefined index: REQUEST_METHOD


daigo75:
# 🐞 bug report

**Behavior**:
A customer reported that the PHP log file contains a large amount of notices indicating that `$_SERVER['REQUEST_METHOD']` is missing. This seems to occur in customer's specific environment, which includes some optimisations that might result in a "not completely standard" environment.

The notice is the following (redacted for brevity):
`PHP Notice: Undefined index: REQUEST_METHOD in /path/to/freemius/wordpress-sdk/config.php on line 166` 

Affected line:
https://github.com/Freemius/wordpress-sdk/blob/84a9be4717effd7697a217e0d931f48ae0d2ecc6/config.php#L166

**Possible solution**
Adding a check to verify that the expected header is set should be sufficient. The code around the affected line already do that, using `isset()`. Examples:
https://github.com/Freemius/wordpress-sdk/blob/84a9be4717effd7697a217e0d931f48ae0d2ecc6/config.php#L155
https://github.com/Freemius/wordpress-sdk/blob/84a9be4717effd7697a217e0d931f48ae0d2ecc6/config.php#L159
https://github.com/Freemius/wordpress-sdk/blob/84a9be4717effd7697a217e0d931f48ae0d2ecc6/config.php#L160

**Versions**: (*)
 -  `Freemius SDK Version:` 2.4.2
 -  `WordPress Version:` unknown, possibly 5.7 or 5.8
 -  `PHP Version:` unknown, possibly 7.4.x

**Plugin / Theme**: (*)
 -  `Name:` Aelia Blacklister for WooCommerce
 -  `Freemius ID:` 8438

**Note**
The plugin itself is not relevant, in this case. The Blacklister is not yet distributed via Freemius, but the framework is loaded anyway, as part of the autoloading mechanism in our solutions.

vovafeldman:
Closed in https://github.com/Freemius/wordpress-sdk/pull/608 (will be released in the upcoming version).

Issue #452, opened 2020-10-12
Undefined index: REQUEST_METHOD in config.php:166


mircobabini:
# 🐞 bug report

**Behavior**:
 -  `What is the issue? (*)`
Undefined index notice in debug.log. PHP Notice:  Undefined index: REQUEST_METHOD in ... [freemius/config.php on line 166](https://github.com/Freemius/wordpress-sdk/blob/master/config.php#L166). This happens when running wordpress by cli/wp-cron.php. Example: `$ php wp-cron.php`
 -  `What is the expected behavior?`
No notice, please :)

**Versions**: (*)
 -  `Freemius SDK Version:`  latest
 -  `WordPress Version:`  latest
 -  `PHP Version:`  any

**Plugin / Theme**: (*)
 -  `Name:`  ininfluent
 -  `Slug:`  ininfluent
 -  `Freemius ID:`  ininfluent

**Additional Information**:
 -  `Browser Type: (*)` ininfluent
 -  `Browser Version: (*)` ininfluent
 -  `OS: (*)` ininfluent
 -  `Stack Traces:`

----
**Note:** `(*)` indicates required information. Without this information, your issue may be auto-closed.

> You can find your Freemius SDK version at `/freemius/start.php`

> You can find your Freemius product ID in the Freemius dashboard: `SETTINGS -> INTEGRATION`.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

timnolte:
This is blowing up the WordPress logs on a lot of sites because I use the WP-CLI via a server-side cron job to run the WP Cron.

andreuka:
same issue here still.

vovafeldman:
Closed in https://github.com/Freemius/wordpress-sdk/pull/608 (will be released in the upcoming version).

PR #608, opened 2023-01-16
[config] Improved the logic that detects whether the context is an HT…

fajardoleo:
…TP request since there are cases in which the HTTP host is available but the request method is not.


vovafeldman (approved):

vovafeldman:
Closes https://github.com/Freemius/wordpress-sdk/issues/513 & https://github.com/Freemius/wordpress-sdk/issues/452 

This is relevant for:
* WP-CLI executions
* WP Cron using server's crontab (`define('DISABLE_WP_CRON', true)`)

PR #611, opened 2023-01-24
[api] [connectivity] Initial commit for the API connectivity test rem…

fajardoleo:
…oval.


includes/class-freemius.php
```diff
@@ -4110,36 +4117,40 @@ function has_api_connectivity( $flush_if_no_connectivity = false ) {
...
-                // API failure.
-                $this->_add_connectivity_issue_message( $pong );
-            }
-
-            if ( $is_connected ) {
-                FS_GDPR_Manager::instance()->store_is_required( $pong->is_gdpr_required );
+            if ( ! $is_active ) {
+                $this->store_connectivity_info( (object) array( 'is_active' => $is_active ), null );
+            } else {
+                $this->_is_on = true;
```
vovafeldman, line 4125: @fajardoleo, don't we need to also store if the opt-in should be on?

includes/class-freemius.php
```diff
@@ -4034,42 +4034,49 @@ private function should_run_connectivity_test( $flush_if_no_connectivity = false
...
-        private function ping( $blog_id = null, $is_gdpr_test = false ) {
-            if ( WP_FS__SIMULATE_NO_API_CONNECTIVITY ) {
-                return false;
+        private function should_turn_fs_on( $is_update = true ) {
+            if ( empty( $this->_plugin->moderation ) ) {
+                return true;
             }
 
-            $version = $this->get_plugin_version();
+            $optin_config = $this->_plugin->moderation['opt_in'];
```
vovafeldman, line 4049: @fajardoleo we can't assume that the developer correctly uses the flag, so we need to validate `$this->_plugin->moderation['opt_in']` is not empty before accessing it.

includes/class-freemius.php
```diff
@@ -4034,42 +4034,49 @@ private function should_run_connectivity_test( $flush_if_no_connectivity = false
...
-                'version'      => $version,
-                'sdk'          => $this->version,
-                'is_admin'     => json_encode( is_admin() ),
-                'is_ajax'      => json_encode( self::is_ajax() ),
-                'is_cron'      => json_encode( self::is_cron() ),
-                'is_gdpr_test' => $is_gdpr_test,
-                'is_http'      => json_encode( WP_FS__IS_HTTP_REQUEST ),
-            );
+            $visibility = $is_update ?
+                $optin_config['updates'] :
```
vovafeldman, line 4059: @fajardoleo similar to the comment above, in this case, we need to validate the array keys exist with `isset()` before accessing.

includes/class-freemius.php
```diff
@@ -17790,6 +17385,15 @@ function opt_in(
 
                 $this->maybe_modify_api_curl_error_message( $result );
 
+                $this->store_connectivity_info(
```
vovafeldman, line 17388: @fajardoleo it looks like the code assumes the error returned from the API is a connectivity issue. What if the error is related to something else?

includes/class-fs-api.php
```diff
@@ -226,26 +228,51 @@ private function _call( $path, $method = 'GET', $params = array(), $retry = fals
...
+                            ( false !== strpos( $result->error->message, 'curl error 35' ) )
+                        ) {
+                            $force_http = true;
+                            $retry      = true;
+                        }
+                    }
+
+                    if ( $retry ) {
+                        if ( $force_http ) {
+                            self::$_options->set_option( 'api_force_http', true, true );
```
vovafeldman, line 259: @fajardoleo I think we also need to call `Freemius_Api_WordPress::SetHttp();`. If I'm right, probably best to create a `toggle_force_http( $is_http )` private helper method.

includes/class-fs-api.php
```diff
@@ -226,26 +228,51 @@ private function _call( $path, $method = 'GET', $params = array(), $retry = fals
...
-                // Log API errors.
-                $this->_logger->api_error( $result );
+            if ( self::is_api_error( $result ) ) {
+                if ( $this->_logger->is_on() ) {
+                    // Log API errors.
+                    $this->_logger->api_error( $result );
+                }
+
+                if ( $force_http ) {
+                    self::$_options->set_option( 'api_force_http', false, true );
```
vovafeldman, line 274: @fajardoleo, similarly, I think we need to update `Freemius_Api_WordPress` to revet back to HTTPS. I guess we need to add `Freemius_Api_WordPress::SetHttps()` as a helper method.

templates/connect.php
```diff
@@ -414,10 +406,7 @@ class="button button-secondary" tabindex="2"><?php fs_esc_html_echo_x_inline( 'S
 			// Set core permission list items.
 			$permissions = array();
 
-			// Add newsletter permissions if enabled.
-			if ( $is_gdpr_required || $fs->is_permission_requested( 'newsletter' ) ) {
-				$permissions[] = $permission_manager->get_newsletter_permission();
-			}
+            $permissions[] = $permission_manager->get_newsletter_permission();
```
vovafeldman, line 409: @fajardoleo let's include it only when `$fs->is_permission_requested( 'newsletter' )`

vovafeldman (changes requested):

includes/class-freemius.php
fajardoleo: @vovafeldman From what I'm seeing, in `/wp-includes/class-wp-http.php`, these are the only possible errors:
- $response = new WP_Error( 'http_request_failed', __( 'A valid URL was not provided.' ) );
- $response = new WP_Error( 'http_request_not_executed', __( 'User has blocked requests through HTTP.' ) );

So the current logic supports both. In case there's a connection, the error returned from the API will be handled like before.

vovafeldman: I see 👍

includes/class-freemius.php
```diff
@@ -17789,10 +17392,29 @@ function opt_in(
...
+                $is_connected = null;
+
+                if ( empty( $license_key ) && $this->is_enable_anonymous() ) {
+                    $this->skip_connection( fs_is_network_admin() );
+
+                    $is_connected = ( ! FS_Api::is_blocked( $result ) );
+                }
+
+                $this->store_connectivity_info(
+                    (object) array( 'is_active' => true ),
```
vovafeldman, line 17406: @fajardoleo, please add a comment above explaining why `'is_actiove'` is `true`, which I assume is because if the user is opting in or activating a license, it means the SDK was turned on for them. Otherwise, this logic would not be executed in the first place.

includes/class-freemius.php
```diff
@@ -24245,6 +23895,20 @@ private function maybe_modify_api_curl_error_message( $result ) {
                 );
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.4
+         *
+         * @param object $result
+         */
+        private function maybe_modify_api_blocked_error_message( $result ) {
```
vovafeldman, line 23904: @fajardoleo I'm all up for code reusability, but since it's only used 2 times and doesn't encapsulate much logic, I think it will be better to use the following instead explicitly:
```php
if ( FS_Api::is_blocked( $result ) ) {
    $result->error->message = $this->generate_api_blocked_notice_message_from_result( $result );
}
```

Also, it will produce fewer lines of code  (even without counting the PHP-Doc).

includes/sdk/FreemiusWordPress.php
```diff
@@ -333,6 +342,30 @@ private static function ExecuteRequest( $pUrl, &$pWPRemoteArgs ) {
...
+            $response = wp_remote_request( $pUrl, $pWPRemoteArgs );
+
+            if (
+                ! is_wp_error( $response ) &&
+                (
+                    empty( $response['headers'] ) ||
+                    empty( $response['headers']['x-api-server'] )
+                )
+            ) {
+                $response = new WP_Error( 'api_blocked', htmlentities( $response['body'] ) );
```
vovafeldman, line 363: Please add a comment explaining the headers check and why we consider it as a blocked API.

includes/class-fs-api.php
```diff
@@ -226,26 +228,49 @@ private function _call( $path, $method = 'GET', $params = array(), $retry = fals
...
-                            ( time() - strtotime( $result->error->timestamp ) ) :
-                            false;
-
-                        // Try to sync clock diff.
-                        if ( false !== $this->_sync_clock_diff( $diff ) ) {
-                            // Retry call with new synced clock.
-                            return $this->_call( $path, $method, $params, true );
+                    $retry = false;
+
+                    if ( ! $in_retry ) {
```
vovafeldman, line 235: I believe this criterion can be moved to the parent `if` (the one starting with `if ( null !== $result &&`), which will remove the unnecessary indentation. It now makes more sense due to the extra `if ( $retry ) {` below, which never gets executed if `$in_retry` is `true`.

includes/class-fs-api.php
```diff
@@ -337,6 +362,25 @@ function get( $path = '/', $flush = false, $expiration = WP_FS__TIME_24_HOURS_IN
 			return $cached_result;
 		}
 
+        /**
+         * This method will be removed in the future after migrating Freemius::safe_remote_post() to FS_Api::call().
```
vovafeldman, line 366: I'd propose changing this to:
`@todo Remove this method after migrating Freemius::safe_remote_post() to FS_Api::call().`

includes/class-fs-api.php
```diff
@@ -617,14 +592,52 @@ static function is_api_error( $result ) {
...
-                isset( $result->error->message )
+                ( $ignore_message || isset( $result->error->message ) )
+            );
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.4
+         *
+         * @param mixed $response
```
vovafeldman, line 611: I think worth replacing `mixed` with `WP_Error|object|string`.

includes/class-fs-api.php
```diff
@@ -617,14 +592,52 @@ static function is_api_error( $result ) {
...
+         *
+         * @return bool
+         */
+        static function is_ssl_error_response( $response ) {
+            $http_error = null;
+
+            if ( $response instanceof WP_Error ) {
+                if (
+                    isset( $response->errors ) &&
+                    isset( $response->errors['http_request_failed'] )
```
vovafeldman, line 621: If I'm not mistaken, you previously mentioned that the WP conventions also include adding space within square brackets, i.e., `$response->errors[ 'http_request_failed' ]`. If true, also applies to the statement within the `if`.

vovafeldman (changes requested):

includes/class-fs-api.php
fajardoleo: @vovafeldman Spaces are used only when the index is a variable:
https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#space-usage

includes/class-freemius.php
```diff
@@ -4034,42 +4034,62 @@ private function should_run_connectivity_test( $flush_if_no_connectivity = false
...
-                'version'      => $version,
-                'sdk'          => $this->version,
-                'is_admin'     => json_encode( is_admin() ),
-                'is_ajax'      => json_encode( self::is_ajax() ),
-                'is_cron'      => json_encode( self::is_cron() ),
-                'is_gdpr_test' => $is_gdpr_test,
-                'is_http'      => json_encode( WP_FS__IS_HTTP_REQUEST ),
-            );
+            if (
+                WP_FS__IS_LOCALHOST_FOR_SERVER &&
```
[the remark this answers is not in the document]
fajardoleo: @vovafeldman Makes sense!

vovafeldman (approved):
Ready for testing ✅ 💪

PR #613, opened 2023-02-11
[api] [connectivity] Improved and fixed an issue with the updating of…

fajardoleo:
… the connectivity info.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -4194,6 +4194,20 @@
         }
 
         /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.4
+         *
+         * @param bool $is_connected
+         */
+        private function update_connectivity_info( $is_connected ) {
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -4193,6 +4193,20 @@ private function store_connectivity_info( $pong, $is_connected ) {
...
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.5.4
+         *
+         * @param bool $is_connected
+         */
+        private function update_connectivity_info( $is_connected ) {
+            $this->store_connectivity_info(
+                // This is true since we update the connection info only after a successful opt-in or license activation which means that Freemius has already been on even before the process.
+                true,
```
vovafeldman, line 4205: @fajardoleo isn't this line should be `(object) array( 'is_active' => true ),`?

PR #618, opened 2023-03-05
Prepare Release 2.5.4

swashata:
We use the internal SDK version as `2.5.4.2` and not `2.5.4` because in
one of the previous releases, we tagged it as `2.5.4.1` and it has been
out in the wild.
    
The next version would automatically simplify it.

### DIFF
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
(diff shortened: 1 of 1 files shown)


fajardoleo (approved):
Looks great to me!

PR #620, opened 2023-03-16
[loader] Ensure that the plugin containing the newest SDK is always l…

fajardoleo:
…oaded first to prevent incompatibility issues with non-standard SDK versions.

### DIFF
--- a/require.php
+++ b/require.php
@@ -44,6 +44,10 @@
 	require_once WP_FS__DIR_INCLUDES . '/entities/class-fs-payment.php';
 	require_once WP_FS__DIR_INCLUDES . '/entities/class-fs-plugin-license.php';
 	require_once WP_FS__DIR_INCLUDES . '/entities/class-fs-subscription.php';
+    /**
+     * Preemptively load the latest version of `Freemius_Api_WordPress` class so that other non-standard integration and custom forks do not override it. For non-standard or custom forks, we recommend copying the source code and renaming the classes, or putting them under namespaces.
+     */
+    require_once WP_FS__DIR_SDK . '/FreemiusWordPress.php';
 	require_once WP_FS__DIR_INCLUDES . '/class-fs-api.php';
 	require_once WP_FS__DIR_INCLUDES . '/class-fs-plugin-updater.php';
 	require_once WP_FS__DIR_INCLUDES . '/class-fs-security.php';
--- a/start.php
+++ b/start.php
@@ -161,9 +161,7 @@
 		);
 	}
 
-    $redirect_after_reordering_plugins = false;
-
-    $is_current_sdk_newest = isset( $fs_active_plugins->newest ) && ( $this_sdk_relative_path == $fs_active_plugins->newest->sdk_path );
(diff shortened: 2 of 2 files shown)


start.php
```diff
@@ -170,7 +172,10 @@
 		fs_update_sdk_newest_version( $this_sdk_relative_path, $fs_active_plugins->plugins[ $this_sdk_relative_path ]->plugin_path );
 
 		$is_current_sdk_newest = true;
-	} else if ( version_compare( $fs_active_plugins->newest->version, $this_sdk_version, '<' ) ) {
+
+        // Try to reorder plugins so that the plugin containing the newest SDK version will be loaded first.
```
swashata, line 176: @fajardoleo So it will mean we intend to redirect for the time, `$fs_active_plugins->newest` is not set. Event if this becomes the only Freemius SDK in the whole plugins/themes installs? If so, then I suggest to enrich the comment.

start.php
```diff
@@ -170,7 +172,10 @@
...
+        $redirect_after_reordering_plugins = ( true === fs_newest_sdk_plugin_first() );
```
swashata, line 177: @fajardoleo I see the signature and code of the function `fs_newest_sdk_plugin_first()`. The return type is `bool`. So is this `( true === fs_newest_sdk_plugin_first() )` comparison necessary?

```php
$redirect_after_reordering_plugins = fs_newest_sdk_plugin_first()
```

swashata (commented):
@fajardoleo Please see my couple of comments. About the white spacing, are we now mixing tabs and spaces in the files? If so, then we are good, if not please change spaces to tabs.

require.php
```diff
@@ -44,6 +44,10 @@
 	require_once WP_FS__DIR_INCLUDES . '/entities/class-fs-payment.php';
 	require_once WP_FS__DIR_INCLUDES . '/entities/class-fs-plugin-license.php';
 	require_once WP_FS__DIR_INCLUDES . '/entities/class-fs-subscription.php';
+    /**
+     * Preemptively load the latest version of `Freemius_Api_WordPress` class so that other non-standard integration and custom forks do not override it. For non-standard or custom forks, we recommend copying the source code and renaming the classes, or putting them under namespaces.
+     */
+    require_once WP_FS__DIR_SDK . '/FreemiusWordPress.php';
```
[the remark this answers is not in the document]
fajardoleo: @drory82 In case you're referring to spaces, we have started to transition from tabs to spaces since a long time ago. However, if you meant something else, would you please clarify?

Tobias-Conrad (approved):
I do not see issues, errors, will keep you updated and i will publish to all my WebinarIgnition (WI) user so WI can not be blamed to be activated and cause issue.

I only added the code to WI, because on disable-admin-notices i could not find the SDK.

Best regards Tobias

https://www.awesomescreenshot.com/video/15673551?key=bb4d56a638ce26edb1fc91e984008b84

vovafeldman:
Due to the high risk of changing the logic in `start.php` (plus we received a report of an infinite redirect of this patched version), we found a more straightforward solution with minimal risks to bypass the incompatibility issue, which is now released as [2.5.5](https://github.com/Freemius/wordpress-sdk/releases/tag/2.5.5).

PR #621, opened 2023-03-19
[backward-compatibility] [api] [php-sdk] Added backward compatibility…

fajardoleo:
… support for logic that relies on outdated PHP SDK.

### DIFF
--- a/includes/class-fs-api.php
+++ b/includes/class-fs-api.php
@@ -378,7 +378,22 @@
                 require_once WP_FS__DIR_SDK . '/FreemiusWordPress.php';
             }
 
-            return Freemius_Api_WordPress::RemoteRequest( $url, $remote_args );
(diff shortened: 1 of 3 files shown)


vovafeldman (approved):

includes/class-fs-api.php
```diff
@@ -462,7 +477,7 @@ private function toggle_force_http( $is_http ) {
 
             if ( $is_http ) {
                 Freemius_Api_WordPress::SetHttp();
-            } else {
+            } else if ( method_exists( 'Freemius_Api_WordPress', 'SetHttps' ) ) {
```
drory82, line 480: Skipping HTTPS as a temp fix is the desired patch ?
For RemoteRequest, you did execute the same logic without calling the new function, here, SetHttps won't be called and that's it. By purpose ?

swashata (approved):
LGTM

PR #624, opened 2023-03-26
[api] [error-handling] When checking a response header, ensure that t…

fajardoleo:
…he response is an array as it can also be a WP_Error instance.

### DIFF
--- a/includes/class-fs-api.php
+++ b/includes/class-fs-api.php
@@ -371,7 +371,7 @@
          * @param string $url
          * @param array  $remote_args
          *
-         * @return mixed
+         * @return array|WP_Error The response array or a WP_Error on failure.
          */
         static function remote_request( $url, $remote_args ) {
             if ( ! class_exists( 'Freemius_Api_WordPress' ) ) {
--- a/includes/class-fs-api.php
+++ b/includes/class-fs-api.php
@@ -386,8 +386,11 @@
             $response = wp_remote_request( $url, $remote_args );
 
             if (
-                empty( $response['headers'] ) ||
-                empty( $response['headers']['x-api-server'] )
+                is_array( $response ) &&
+                (
+                    empty( $response['headers'] ) ||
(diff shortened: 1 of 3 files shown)


includes/sdk/FreemiusWordPress.php
```diff
@@ -354,8 +354,11 @@ static function RemoteRequest( $pUrl, $pWPRemoteArgs ) {
             $response = wp_remote_request( $pUrl, $pWPRemoteArgs );
 
             if (
-                empty( $response['headers'] ) ||
-                empty( $response['headers']['x-api-server'] )
+                is_array( $response ) &&
```
swashata, line 357: @fajardoleo The `wp_remote_request` is guaranteed to return either an `array` or `WP_Error`. This is what we are doing in this method (+ the replica method in `class-fs-api.php`) too.

So I propose to just enhance the `@return` of this function and also `FS_Api#remote_request` to `@return array|WP_Error`.

Now about this `is_array` check, I was thinking if `!is_wp_error( $response )` is more semantic here. Please take a 2nd look.

includes/class-fs-api.php
```diff
@@ -386,8 +386,11 @@ static function remote_request( $url, $remote_args ) {
             $response = wp_remote_request( $url, $remote_args );
 
             if (
-                empty( $response['headers'] ) ||
-                empty( $response['headers']['x-api-server'] )
+                is_array( $response ) &&
```
swashata, line 389: @fajardoleo Please see below.

swashata (commented):
@fajardoleo Please see my small suggestion.

includes/sdk/FreemiusWordPress.php
fajardoleo: @swashata Thanks! The PHP documentation comment has been updated.

swashata (approved):
Thanks @fajardoleo LGTM.

PR #627, opened 2023-04-04
[i18n] [build] [styles] Update and fix of gulp build.

bahiirwa:
- [x] Fix: Update the gulp packages and migrate tasks runners to gulp 4.
- [x] Fix: Use of calc() to fix deprecated SCSS division rules.
- [x] Fix: Compilation of .pot file.
- [x] Add: New rules to package.json for local style builds and potential merge/tag build automation.


gulpfile.js
```diff
@@ -106,39 +112,80 @@ gulp.task('translations-feelup', ['prepare-translations'], function () {
...
+	];
+
+	// location of style
+	return gulp.src( './assets/scss/**/*.scss' )
+		// Compile file
+		.pipe( sass() )
+		// Use postcss
+		.pipe( postcss( plugins ) )
+		// Push build
+		.pipe( gulp.dest( './assets/css/admin' ) )
```
swashata, line 154: @bahiirwa this will create files inside `assets/css/admin/`. So some files will be nested as `assets/css/admin/admin`, like you can see in the diff of the PR. Please target `assets/css`.

swashata (changes requested):
Hi @bahiirwa great effort 👏 . I have noticed one immediate issue which I have commented. Additionally please modify the feature branch to base it on `develop` and kindly send a PR to `develop`.

gulpfile.js
bahiirwa: Thanks for spotting this. I had two rules for compiling the admin css separately from the front end. Seems, I forgot to refactor this when I joined the two rules.
Fixed it.

gulpfile.js
```diff
@@ -19,10 +19,13 @@ function getFolders(dir) {
...
-var transifex = require('gulp-transifex').createClient(options);
+const options         = require('./transifex-config.json');
+const transifex       = require('gulp-transifex').createClient(options);
+const languagesFolder = './languages/';
+const folders         = getFolders(languagesFolder);
 
 // Create POT out of PHP files
-gulp.task('prepare-source', function () {
+function prepare_source() {
     gulp.src('**/*.php')
```
swashata, line 29: In new gulp we are supposed to return a promise or promise-like to properly signal a task completion. Please read documentation here https://gulpjs.com/docs/en/getting-started/async-completion

So, for all gulp task-related functions, you need to return the stream. 

```js
function foo() {
  return gulp.src(...).pipe(...);
}
```

Then gulp will correctly hint to us about the time it took.

When there are no streams to return, you have various options. I find using the built-in Promise to be very powerful and flexible.

```js
return new Promise((resolve, reject) => {
  // doSomething is an async task, but it is implemented using a callback pattern.
  doSomething(() => {
    resolve();
  });
});
```

You can also use the `reject` to tell gulp the task has failed.

gulpfile.js
```diff
@@ -71,33 +74,36 @@ gulp.task('prepare-source', function () {
         }))
         .pipe(rename('freemius-en.po'))
         .pipe(gulp.dest(languagesFolder));
-});
+}
 
 // Push updated po resource to transifex.
-gulp.task('update-transifex', ['prepare-source'], function () {
+function update_transifex() {
+    prepare_source();
```
swashata, line 81: The `prepare_source` is an async function. It will need some time to complete. Given that this (`prepare_source`) async function is properly implemented, we need to wait for it, before pushing the resource, else there will always be race conditions. (An old version of `freemius-en.po` will be pushed).

Since `prepare_source` is a gulp task, we can use the gulp `series` composition.

```js
function prepare_source() {
  return gulp.src(...)...
}

function push_transifex() {
  return gulp.src(languagesFolder + 'freemius-en.po').pipe(transifex.pullResource());
}

const update_transifex = series(prepare_source, push_transifex);
```

Please pardon me for any interface/syntactical mistakes I might have made.

gulpfile.js
```diff
@@ -71,33 +74,36 @@ gulp.task('prepare-source', function () {
...
     return gulp.src(languagesFolder + 'freemius-en.po')
         .pipe(transifex.pushResource());
-});
+}
 
 // Download latest *.po translations.
-gulp.task('download-translations', ['update-transifex'], function () {
+async function download_translations() {
+    update_transifex();
```
swashata, line 88: @bahiirwa This and all other direct usages of async tasks need refactoring as above.

gulpfile.js
```diff
@@ -106,39 +112,79 @@ gulp.task('translations-feelup', ['prepare-translations'], function () {
...
+
+    return gulp.src( './assets/scss/**/*.scss' )
+        .pipe( sass() )
+        .pipe( postcss( plugins ) )
+        .pipe( gulp.dest( './assets/css/' ) )
+}
+
+// Compile css only in dev mode.
+function watch() {
+    gulp.watch( './assets/scss/*.scss', style );
```
swashata, line 156: @bahiirwa You can use the blob `./assets/scss/**/*.scss` instead of the two `watch`.

gulpfile.js
```diff
@@ -106,39 +112,79 @@ gulp.task('translations-feelup', ['prepare-translations'], function () {
...
-    gulp.run('translations-feelup');
-    gulp.run('cleanup');
-    gulp.run('compile-translations');
-});
\ No newline at end of file
+}
+
+// Run postcss processing for styles.
+function style() {
+    let plugins = [
```
swashata, line 144: @bahiirwa Please use `const`

gulpfile.js
```diff
@@ -106,39 +112,79 @@ gulp.task('translations-feelup', ['prepare-translations'], function () {
...
+exports.update_transifex      = update_transifex;
+exports.download_translations = download_translations;
+exports.prepare_translations  = prepare_translations;
+exports.translations_feelup   = translations_feelup;
+exports.cleanup               = cleanup;
+exports.compile_translations  = compile_translations;
+exports.style                 = style;
+exports.watch                 = watch;
+exports.build                 = build;
+exports.default               = series(
```
swashata, line 182: @bahiirwa I would add the `style` task here too and optimize them, running tasks in parallel which can be run. Like (just an example)

```js
parallel(
  style,
  series(prepare_source, update_transifex, ...)
);
```

swashata: @bahiirwa Thinking about it again 🤔 I would rather not have transifex in `watch` and `build`, instead would have them in a separate `npm run translation` command.

The reason is, for most day-to-day tasks, we just need the style watcher and maybe in the future a JS uglifier (which we can add when needed). So I propose to 

1. Have `npm run dev` and `npm run build` deal with only style for now.
2. Create a separate command for explicitly extracting translation. This will just create/update the `pot` file.
3. Create a separate command for explicitly updating existing `.po` with the `pot` file.
4. Create a separate command for uploading/syncing translations with transifex.
5. Create a separate command to do the above 3, in series.

Please feel free to modify the tasks/order as you see fit. I am yet to dig deep into exactly how the translations are being uploaded/handled. I will do that once we are done with the changes required in the new gulp tasks.

package.json
```diff
@@ -9,20 +9,28 @@
   "main": "gulpfile.js",
   "dependencies": {},
   "scripts": {
+    "dev": "gulp watch",
```
swashata, line 12: @bahiirwa Once we will be done with the PR review, and once this PR will be ready for merge, please create a `CONTRIBUTING.md` file and document how to use the new build system, and what to do for transifex etc... I am sure your documentation expertise will make it easy for new contributors ;)

swashata (changes requested):
@bahiirwa Great job. We are getting there. Please see my comments for the new SCSS and gulp syntax. Thanks.

PR #628, opened 2023-04-19
[translation] Fixed broken translations.

fajardoleo:
Temporary manual fixes for broken translations due to outdated compiler and changes in Transifex .po file version. We will fix the automation from PR #627.


bahiirwa (approved):
Consistent change. Looks Good.

PR #631, opened 2023-05-15
Prepare release 2.5.8

swashata:
- [x] Use `esc_url` instead of `esc_attr` in the opt-in UI template.
- [x] Bump the version in `start.php`.


vovafeldman (approved):
Short and sweet 🍭

PR #633, opened 2023-05-16
[ajax] [url] Fixed the AJAX URL generation so that the URL params are…

fajardoleo:
… properly added.

### DIFF
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.5.8';
+	$this_sdk_version = '2.5.8.1';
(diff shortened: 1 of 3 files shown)


vovafeldman (changes requested):
@fajardoleo, to make it more resilient, it's better to set it in a local var:
`var ajaxUrl = <?php echo Freemius::ajax_url() ?>;`

And then `url: ajaxUrl + (ajaxUrl.includes('?') ? '&' : '?') + ...`

This way even if we change `Freemius::ajax_url()` it will still work properly.

vovafeldman (approved):

PR #634, opened 2023-05-18
[tabs] [affiliation] When the current tab is the affiliation tab, mak…

fajardoleo:
…e sure that saved tabs are properly added.


vovafeldman (approved):
@fajardoleo

PR #639, opened 2023-06-05
[clone-resolution] [site-url] Fixed the retrieving of the site URL so…

fajardoleo:
… that it will be retrieved properly when it's set in the wp-config.php file.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -3492,6 +3492,28 @@
          * @return string
          */
         static function get_unfiltered_site_url( $blog_id = null, $strip_protocol = false, $add_trailing_slash = false ) {
+            $url = ( ! is_multisite() && defined( 'WP_SITEURL' ) ) ? WP_SITEURL : self::get_site_url_from_wp_option( $blog_id );
+
+            if ( $strip_protocol ) {
+                $url = fs_strip_url_protocol( $url );
+            }
+
+            if ( $add_trailing_slash ) {
+                $url = trailingslashit( $url );
+            }
+
+            return $url;
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -3492,29 +3492,35 @@ function is_clone( $only_if_manual_resolution_is_not_hidden = false ) {
          * @return string
          */
         static function get_unfiltered_site_url( $blog_id = null, $strip_protocol = false, $add_trailing_slash = false ) {
-            global $wp_filter;
+            $url = defined( 'WP_SITEURL' ) ?
```
swashata, line 3495: @fajardoleo Maybe it would cleaner to write the logic like this

```php
$url = defined('WP_SITEURL') ? WP_SITEURL : self::get_site_url_from_wp_option( $blog_id );

if ( $strip_protocol ) {
  // ...
}
// ...
```

And extract the big logic with global `$wp_filter` to a helper function.

I wonder @fajardoleo , should we only rely on `WP_SITEURL` when `$blog_id` is `null`?

swashata (changes requested):
@fajardoleo Please see my question and a small suggestion. Thanks.

includes/class-freemius.php
fajardoleo: Thanks @swashata. I have refactored the code. With regard to `$blog_id == null`, I noticed that the relevant constants are actually ignored by the core code when the context is multisite (https://github.com/WordPress/WordPress/blob/703d5bdc8deb17781e9c6d8f0dd7e2c6b6353885/wp-includes/ms-default-filters.php#L83), so I added the relevant condition instead.

swashata (approved):
Looks good now @fajardoleo

PR #626, opened 2023-03-28
[account] [addon] [license] [php-8.1] Fixed an issue with the expirat…

fajardoleo:
…ion label of lifetime add-on licenses.

PR #647, opened 2023-07-24
Fix of strtotime(): Passing null to parameter #1 ($datetime) of type string is deprecated

vovafeldman:
[templates] [addon] [bug-fix] The `expiration` prop of a lifetime license is `null`, so `strtotime()` is throwing a PHP error for that case. Therefore, I moved the string generation logic into the `if` statement, ensuring the license is not lifetime and is associated with an active subscription.

### DIFF
--- a/templates/account/partials/addon.php
+++ b/templates/account/partials/addon.php
@@ -240,19 +240,19 @@
                         true
                     );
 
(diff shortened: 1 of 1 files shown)


vovafeldman:
Closing as this one is duplicate 🤦‍♂️ https://github.com/Freemius/wordpress-sdk/pull/626

PR #649, opened 2023-08-09
[garbage-collector] Initial commit.

fajardoleo:


includes/class-freemius.php
```diff
@@ -425,6 +425,14 @@ private function __construct( $module_id, $slug = false, $is_init = false ) {
 
             $this->_storage = FS_Storage::instance( $this->_module_type, $this->_slug );
 
+            // If not set or 24 hours have already passed from the last time it's set, set the last load timestamp to the current time.
```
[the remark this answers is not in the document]
swashata: Thinking about it again @fajardoleo, maybe it won't be good to do it in the ctor of `FS_Storage`. We might load different "storage" data which won't need it.

In any case, I think just setting the `last_load_timestamp` won't really save it right? So how about you also save it.

```php
$this->_storage->last_load_timestamp = time();
$this->_storage->save();
```

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,318 @@
...
+            }
+
+            $this->clean_up();
+
+            // Create a 1-day lock.
+            $lock->lock( WP_FS__TIME_24_HOURS_IN_SEC );
+        }
+
+        private function clean_up() {
+            $product_types = $this->get_product_types();
```
swashata, line 61: @fajardoleo This function is a bit difficult and complex to go through and understand :)

I propose to refactor it like this


```php
private function clean_up() {
  $_accounts = FS_Options::instance( WP_FS__ACCOUNTS_OPTION_NAME, true );
  $products_to_delete = array();

  foreach ( $this->get_product_types() as $product_type ) {
    $products_to_delete = array_merge( 
       $products_to_delete, 
       $this->get_products_to_delete( $product_type, $_accounts ) 
    );
  }

  if ( ! empty( $products_to_delete ) ) {
    $this->delete_products( $products_to_delete, $_accounts );
    $this->delete_inactive_users( $_accounts );
  }
}

private function get_products_to_delete( $product_type, $accounts ) {
  $option_name      = ( $product_type . 's' );
  $all_product_data = $_accounts->get_option( $option_name, array() );

  $products_to_delete = array();

  foreach ( $all_product_data as $slug => $product_data ) {
    if ( $this->is_product_active( $slug, $product_data ) ) {
      continue;
    }

    // do your other stuff, but don't delete anything yet.
    // I think we can return just the `slug => $product_data` pair here.
  }

  return $products_to_delete;
}

private function delete_products( $products_to_delete, $_accounts ) {
  // Rearrange so that all add-ons are on top.

  // Delete the products + associated options. You can break this function down too.

  // Don't delete the users yet.
}

private function delete_inactive_users( $_accounts ) {
  // Loop and find all users.
  // Delete the users without any "install".
}
```

Another approach could be to break down this entire logic into multiple "strategy" classes. Statetiges would involve "how to delete inactive plugins", "how to delete inactive themes" etc. But I leave it to you, I don't think it is very important.

But please do see if/how we can simplify the big logic inside this function. It is very hard to understand and would be very difficult to debug. I would urge you to explore logic/algorithm where we would loop multiple times but the code would be readable and understandable.

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,318 @@
...
+            $has_updated_option = false;
+
+            foreach( $products_slugs_by_type as $product_type => $products ) {
+                foreach( $products as $product ) {
+                    $product_id   = $product['id'];
+                    $slug         = $product['slug'];
+                    $product_file = $product['file'];
+
+                    if ( ! $loaded_products_options_by_type[ $product_type ] ) {
+                        $products_data_by_type[ $product_type ]           = $this->load_options( $products_options_by_type[ $product_type ] );
```
swashata, line 139: @fajardoleo aren't you overriding what was set on line `116`. Do we really need both?

swashata (changes requested):
@fajardoleo Thanks for the PR. I think I understand the logic, but I do believe the code can be made more organized. Please see my suggestion, thanks 👍

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,359 @@
...
+
+            $products_to_delete = array();
+
+            foreach ( $this->get_product_types() as $product_type ) {
+                $products_to_delete[ $product_type ] = $this->get_products_to_delete( $_accounts, $product_type );
+            }
+
+            $update_options = false;
+
+            if ( ! empty( $products_to_delete ) ) {
```
swashata, line 68: @fajardoleo Continued from below...

If we do it like this

```php
if ( empty( $products_to_delete ) ) {
  return;
}
```

Then we won't need to hold a flag and compromise on performance. Plus I think the code would be cleaner.

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,359 @@
...
+
+                $is_addon = FS_Plugin::is_valid_id( $product_data->parent_plugin_id );
+
+                if ( ! $is_addon ) {
+                    $products_to_delete[] = $product_data;
+                } else {
+                    /**
+                     * If add-on, add to the beginning of the array so that add-ons are removed before their parent. This is to prevent an unexpected issue when an add-on exists but its parent was already removed.
+                     */
+                    array_unshift( $products_to_delete, $product_data );
```
swashata, line 124: @fajardoleo If a theme has add-ons, this logic might not hold and by your comment, it can still produce issues.

So I propose to completely separate this step. Instead of doing it here, you can do the rearranging inside the `delete_products` steps where it will iterate over all products and slice the add-ons and the rearrange by putting them all on top.

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,359 @@
...
+        }
+
+        /**
+         * @param FS_Options $_accounts
+         * @param array[]    $products_to_delete
+         *
+         * @return bool
+         */
+        private function delete_products( $_accounts, $products_to_delete ) {
+            $products_option_names_by_type       = $this->get_product_option_names_by_type();
```
swashata, line 163: @fajardoleo I would remove this variable (more suggestion continued).

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,359 @@
...
+            $products_options_by_type            = array();
+            $has_loaded_products_options_by_type = array_fill_keys( array_keys( $products_to_delete ), false );
```
swashata, line 165: @fajardoleo this won't be needed (contd...)

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,359 @@
...
+
+            $has_updated_option = false;
+
+            foreach( $products_to_delete as $product_type => $products ) {
+                foreach( $products as $product ) {
+                    $slug = $product->slug;
+
+                    if ( ! $has_loaded_products_options_by_type[ $product_type ] ) {
```
swashata, line 173: @fajardoleo continuing from the above two suggestion, here you can do

```php
if ( ! isset($products_options_by_type[ $product_type ])) {
  $products_options_by_type[ $product_type ] = $this->load_options($_accounts, $this->get_product_option_names( $product_type );
}
```

So:
1. Rename/Refactor `get_product_option_names_by_type` to `get_product_option_names`.
2. The function will take the `$product_type` as param and would return the relevant options.

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,359 @@
...
+                    }
+                }
+            }
+
+            return $has_updated_option;
+        }
+
+        /**
+         * @param FS_Options $_accounts
+         * @param array      $installs_count_by_user_id
```
swashata, line 229: @fajardoleo Please update the php-doc.

swashata (changes requested):
@fajardoleo Great job on the refactoring 👏 . Please see my few suggestions, I think we can simplify it/make it more readable. Thanks

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,458 @@
...
+            $this->_accounts = $_accounts;
+            $this->_types    = $types;
+        }
+
+        public function clean() {
+            $users = Freemius::get_all_users();
+
+            $user_has_install = array();
+
+            foreach ( $this->_types as $product_type ) {
```
swashata, line 361: @fajardoleo Proposing to extract this to a private function for readability and rename it to `$user_has_install_map` or something like that:

```
$user_has_install_map = $this->get_user_has_install_map();
```

swashata (changes requested):
@fajardoleo Nice refactoring. You have managed to get rid of several `if..else` in the logic and the code is much more readable now. Please see my suggestion for some easy improvements. Thanks.

includes/class-fs-garbage-collector.php
```diff
@@ -0,0 +1,458 @@
...
+                            }
+                        }
+
+                        if ( ! $updated ) {
+                            continue;
+                        }
+                    }
+
+                    if ( ! $updated ) {
+                        if ( isset( $option[ $slug ] ) ) {
```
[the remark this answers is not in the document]
swashata: Thanks @fajardoleo looks like I get it now. If we ever go with my improvement suggestion (one GC class per option) we can get rid of this, but not important right now at the moment. I have added a `@todo`.

swashata (commented):
@fajardoleo I implemented my suggestions and fixed the typos I found. Now please review my commit and test :D

swashata (approved):
LGTM

PR #651, opened 2023-08-23
Deployment 23 August, 2023 - Version 2.5.11

swashata:
**Version 2.5.11**

## Fixes in clone resolution and multisite license activation

Version 2.5.11 brings in the following bug fixes:

1. Properly acknowledge the use of the constant `WP_SITEURL` so that the clone resolution mechanism can work properly in some edge cases.
2. Fixed multisite license activation mechanism where in some cases it would not include inactive sites.

Additionally, we have done some housecleaning. 

1. We implemented PHPStan for static analysis to eagerly encounter zero-day bugs. 
2. PHPCS' PHPCompatibility was also added to make sure we are supporting PHP versions from `5.6` to the latest.


fajardoleo (approved):
Looks good to me.

PR #652, opened 2023-08-23
[fix] [backward-compatibility] [server-compatibility] Safely use the …

swashata:
…new idn_to_ascii functions only when the system constants are available.


fajardoleo (approved):
Looks good to me.

PR #655, opened 2023-08-31
[affiliation] Fixed the retrieving of the default email address.

fajardoleo:


includes/class-freemius.php
```diff
@@ -14793,7 +14793,9 @@ function _submit_affiliate_application() {
             if ( ! $this->is_registered() ) {
                 // Opt in but don't track usage.
                 $next_page = $this->opt_in(
-                    false,
```
swashata, line 14796: @fajardoleo I think it is worth doing a sanity check here. Also from the form and flow it looks like if the user isn't opted in, we better get the email from the form.

```php
if ( ! $this->is_registered() ) {
  $email = isset( $affiliate['email'] ) ? $affiliate['email'] : '';

  if ( ! self::is_valid_email( $email ) ) {
    self::shoot_ajax_failure('Invalid email address');
  }

  // ...
}

**Review by swashata — CHANGES_REQUESTED** [review]
@fajardoleo Please see my one suggestion, thanks.

File: start.php [hunk 2]
```diff
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.5.11';
+	$this_sdk_version = '2.5.12';
```
swashata, line 18: @fajardoleo Rebase this branch please and bump the version by our convention. `2.5.12.1` if I am not mistaken?

swashata (changes requested):
@fajardoleo Please see the one last request about bumping the version and will be done. Thanks.

swashata (approved):
LGTM

PR #656, opened 2023-08-31
[code-type-change] Also handle code type changes for anonymous users.

fajardoleo:


includes/class-freemius.php
```diff
@@ -5644,14 +5644,14 @@ function _after_code_type_change() {
...
-                }
+            if ( ! $this->is_addon() ) {
+                add_action(
+                    is_admin() ? 'admin_init' : 'init',
+                    array( &$this, '_plugin_code_type_changed' )
+                );
+            }
 
+            if ( $this->is_registered() ) {
                 if ( $this->is_premium() ) {
```
swashata, line 5655: @fajardoleo We can simplify `if ( $this->is_registered() && $this->is_premium() )`

swashata (changes requested):
@fajardoleo Changes look good. Please see my one small suggestion.

swashata (approved):
LGTM

PR #658, opened 2023-09-03
Deployment, Version 2.5.12

swashata:
## Fixed update notice issue when free and premium versions of a plugin is running in parallel

We discovered an edge case in our SDK when a plugin's free and premium version needs to be running in parallel. In some cases, the updates from the premium version would show up under the Free version, causing confusion.

The SDK v2.5.12 fixes the issue.


fajardoleo (approved):
Looks good to me.

PR #660, opened 2023-09-19
[sdk-loader] [themes] Fixed a conflict between the SDK loader and the…

fajardoleo:
… theme live preview functionality.

### DIFF
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.5.12';
+	$this_sdk_version = '2.5.12.1';
 
 	#region SDK Selection Logic --------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


start.php
```diff
@@ -46,6 +46,12 @@
 	 */
 	$file_path    = fs_normalize_path( __FILE__ );
 	$fs_root_path = dirname( $file_path );
+
+    if ( ! function_exists( 'wp_get_current_user' ) ) {
```
swashata, line 50: @fajardoleo This could be a bit risky, since it will load `pluggable` even when not needed, also in the front-end. Also, this might load the `pluggable` eagerly possibly breaking some other plugin that wants to override the functions.

How about we do it like this?

```php
    if (
        ! function_exists( 'wp_get_current_user' ) &&
        /**
         * `get_stylesheet` will rely on `wp_get_current_user` when it is being filtered by `theme-previews.php`. That happens only when the site editor is loaded or when the site-editor is sending REST requests.
         * @see theme-previews.php:wp_get_theme_preview_path()
         * @todo If this behavior is fixed in the core, we will remove this workaround.
         * @since WP 6.3.0
         */
        (
            'site-editor.php' === basename( $_SERVER['SCRIPT_FILENAME'] ) ||
            ( wp_is_json_request() && ! empty( $_GET['wp_theme_preview'] ) )
        )
    ) {
        // Requiring this file since the call to get_stylesheet() below can trigger a call to wp_get_current_user() when previewing a theme.
        require_once ABSPATH . 'wp-includes/pluggable.php';
    }
```

I saw the error happening only when live previewing "Block Themes", so adjusted the code accordingly.

swashata (changes requested):
Please see suggestion, thanks.

swashata (approved):
Great 👍

PR #661, opened 2023-09-19
[tabs] [php-8] Ensure that there's a valid stored tabs HTML string be…

fajardoleo:
…fore attempting to process it to prevent triggering a deprecated usage notice.


swashata (approved):
LGTM

PR #663, opened 2023-09-26
[url] [php-8] Improved the URL parsing logic to prevent triggering wa…

fajardoleo:
…rnings from PHP 8.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -3123,11 +3123,8 @@
                 return false;
             }
 
-            $url_params = array();
-            parse_str( parse_url( $url, PHP_URL_QUERY ), $url_params );
-
-            $sub_url_params = array();
-            parse_str( parse_url( $sub_url, PHP_URL_QUERY ), $sub_url_params );
+            $url_params     = fs_parse_url_params( $url );
+            $sub_url_params = fs_parse_url_params( $sub_url );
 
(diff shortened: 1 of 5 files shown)


includes/class-freemius.php
```diff
@@ -3123,11 +3123,19 @@ private function is_matching_url( $sub_url, $url = '' ) {
...
             }
 
-            $url_params = array();
-            parse_str( parse_url( $url, PHP_URL_QUERY ), $url_params );
+            $url_components = parse_url( $url, PHP_URL_QUERY );
+            $url_params     = array();
 
-            $sub_url_params = array();
-            parse_str( parse_url( $sub_url, PHP_URL_QUERY ), $sub_url_params );
+            if ( ! empty( $url_components ) ) {
```
swashata, line 3129: @fajardoleo an immediate improvement we can do here is to extract the common logic in a function. Perhaps a dedicated `fs_parse_url_params` function in the `fs-core-functions`.

So it would become

```php
$url_params = fs_parse_url_params( $url );
$sub_url_params = fs_parge_url_params( $sub_url );
```

swashata (changes requested):
@fajardoleo Please see my one suggestion, looks good.

swashata (approved):
LGTM

PR #669, opened 2023-11-27
[start-loader] [fix] Remove temporary patch for a bug in the gutenberg…

swashata:
… core which was adding filters to stylesheets that relied on pluggable functionalities, causing it to fail.


fajardoleo (approved):
Looks good to me.

Issue #675, opened 2023-12-12
White space on top of admin because of unseen PHP error


bfintal:
# 🐞 bug report

**Behavior**:
 -  `What is the issue? (*)`
In the admin, I see this white space at the top, which is caused by a PHP error coming from the Freemius SDK.

<img width="554" alt="image" src="https://github.com/Freemius/wordpress-sdk/assets/1033611/f6f82232-096f-452b-8ac2-71b7350c2f29">

 If you print the last error in PHP, you'll see this error:

```
Creation of dynamic property FS_User::$is_beta is deprecated
```

 -  `What is the expected behavior?`
 No error, no white space

PR #676, opened 2023-12-12
fix: added missing $is_beta class property to FS_User

bfintal:
Fixes https://github.com/Freemius/wordpress-sdk/issues/675

### DIFF
--- a/includes/entities/class-fs-user.php
+++ b/includes/entities/class-fs-user.php
@@ -38,6 +38,10 @@
 		 * @var float
 		 */
 		public $gross;
+		/**
+		 * @var bool
+		 */
+		public $is_beta;
 
 		#endregion Properties
 


swashata:
Hey @bfintal thank you for the PR. As I can see it is adding a property to the `FS_User` class. But we have moved the `is_beta` from user to site/install quite some time back.

So could it be that the error you're getting is due to some old/legacy account data? We have also incorporated a [garbage collector](https://github.com/Freemius/wordpress-sdk/releases/tag/2.6.2), so that old data might have gotten deleted already.

Could you please confirm if this fix is still needed, even after running the garbage collector?

andreiglingeanu:
@swashata I can confirm that removing the cached serialized objects from the DB helped fix this issue in my end.
So yes, indeed adding this flag in the class is not needed from my point of view.

PR #687, opened 2024-02-11
[wordpress-compliance] Fix the loading/enqueueing of JS scripts in WP.

bahiirwa:
- [x] Fix the loading/enqueueing of JS scripts in WP.. The default is false to load in the head. Ref: https://developer.wordpress.org/reference/functions/wp_enqueue_script/#parameters

### DIFF
--- a/includes/fs-core-functions.php
+++ b/includes/fs-core-functions.php
@@ -118,7 +118,7 @@
     }
 
     if ( ! function_exists( 'fs_enqueue_local_script' ) ) {
-        function fs_enqueue_local_script( $handle, $path, $deps = array(), $ver = false, $in_footer = false ) {
+        function fs_enqueue_local_script( $handle, $path, $deps = array(), $ver = false, $in_footer = true ) {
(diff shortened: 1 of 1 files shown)


includes/fs-core-functions.php
```diff
@@ -118,7 +118,7 @@ function fs_enqueue_local_style( $handle, $path, $deps = array(), $ver = false,
     }
 
     if ( ! function_exists( 'fs_enqueue_local_script' ) ) {
-        function fs_enqueue_local_script( $handle, $path, $deps = array(), $ver = false, $in_footer = 'all' ) {
+        function fs_enqueue_local_script( $handle, $path, $deps = array(), $ver = false, $in_footer = false ) {
```
swashata, line 121: @bahiirwa let's change the `$in_footer` default value to `true`. This way we won't be doing any breaking change.

swashata, line 121: Let's change the `$in_footer` default to `true`. Since `all` evaluates to `true`. This way we won't make any breaking change.

swashata (changes requested):
Great catch. Please see my comment. Thanks.

swashata (approved):

PR #691, opened 2024-03-11
[debug] [logging] Automatically turn off `fs_debug_mode`after 24 hours

DanieleAlessandra:
When the user activates debugging a single event cron starts:
`wp_schedule_single_event(time() + 24 * HOUR_IN_SECONDS, 'fs_debug_turn_off_logging_hook');`

When the user deactivates debuggong this cron is unscheduled:
`wp_unschedule_event($timestamp, 'fs_debug_turn_off_logging_hook');`

If the event happens, the option is set to off:
`add_action('fs_debug_turn_off_logging_hook', array( self::class, '_turn_off_debug_mode' ) );`

Also, in the debug page I added a countdown timer, so the user is informed about the fact that debugging will be disabled automatically.
image

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -1257,7 +1257,7 @@
          *
          * @return bool
          */
-        private static function set_network_upgrade_mode( FS_Storage $storage ) {
+        public static function set_network_upgrade_mode( FS_Storage $storage ) {
             return $storage->is_network_activation = true;
         }
 
@@ -1583,6 +1583,8 @@
             ) {
                 add_action( 'admin_init', array( &$this, 'connect_again' ) );
             }
+
+            FS_DebugManager::register_hooks();
         }
 
         /**
@@ -2701,7 +2703,7 @@
          * @author Vova Feldman (@svovaf)
          * @since  2.4.3
          */
-        private static function reset_deactivation_snoozing( $period = 0 ) {
+        public static function reset_deactivation_snoozing( $period = 0 ) {
             $value = ( 0 === $period ) ? null : 'true';
 
             if ( ! is_multisite() || fs_is_network_admin() ) {
@@ -3418,20 +3420,7 @@
 
             self::$_global_admin_notices = FS_Admin_Notices::instance( 'global' );
 
-            if ( ! WP_FS__DEMO_MODE ) {
-                add_action( ( fs_is_network_admin() ? 'network_' : '' ) . 'admin_menu', array(
-                    'Freemius',
-                    '_add_debug_section'
-                ) );
-            }
-
-            add_action( "wp_ajax_fs_toggle_debug_mode", array( 'Freemius', '_toggle_debug_mode' ) );
-
-            self::add_ajax_action_static( 'get_debug_log', array( 'Freemius', '_get_debug_log' ) );
-
-            self::add_ajax_action_static( 'get_db_option', array( 'Freemius', '_get_db_option' ) );
-
-            self::add_ajax_action_static( 'set_db_option', array( 'Freemius', '_set_db_option' ) );
+            FS_DebugManager::load_required_static();
 
             if ( 0 == did_action( 'plugins_loaded' ) ) {
                 add_action( 'plugins_loaded', array( 'Freemius', '_load_textdomain' ), 1 );
@@ -3456,6 +3445,14 @@
             }
 
             self::$_statics_loaded = true;
+        }
+
+        public static function get_static_logger() {
+            return self::$_static_logger;
+        }
+
+        public static function get_accounts() {
+            return self::$_accounts;
         }
 
         #--------------------------------------------------------------------------------
@@ -3626,7 +3623,7 @@
          *
          * @since 2.1.3
          */
-        private static function migrate_options_to_network() {
+        public static function migrate_options_to_network() {
             self::migrate_accounts_to_network();
 
             // Migrate API options from site level to network level.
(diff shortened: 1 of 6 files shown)


templates/debug.php
```diff
@@ -30,6 +32,9 @@
     <div class="fs-switch fs-round <?php echo WP_FS__DEBUG_SDK ? 'fs-on' : 'fs-off' ?>">
         <div class="fs-toggle"></div>
     </div>
+
+    <span class="auto-off-debug-countdown hidden"><?php echo fs_esc_html_echo_x_inline( 'Auto off in:', 'timer for auto-disabling debug' ); ?> <span class="time">23:59:59</span>
```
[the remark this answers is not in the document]
DanieleAlessandra: Yes, since this span is hidden by default there is no need to write the correct value on page load

includes/class-fs-storage.php
swashata, line 1: Please revert the white-space change

includes/managers/class-fs-debug-manager.php
swashata, line 1: Please fix the indentation. We use 4 spaces

templates/debug.php
```diff
@@ -39,18 +43,69 @@
...
                         _wpnonce   : <?php echo wp_json_encode( wp_create_nonce( 'fs_toggle_debug_mode' ) ); ?>,
-                        is_on : ($(this).hasClass( 'fs-on' ) ? 1 : 0)
-                    }, function ( response ) {
+                        is_on
+                    }, function (response) {
+                        if (is_on) {
+                            startCountdownManually();
+                        } else {
+                            stopCountdownManually();
+                        }
```
swashata, line 58: Let's add a new line after this line.

swashata (changes requested):
Hey @DanieleAlessandra nicely done 👏 ... Please see my few comments. Also kindly do a self review and fix the indentation glitches. We are getting close to merging this.

includes/class-fs-storage.php
```diff
@@ -26,6 +26,10 @@ class FS_Storage {
          * @var FS_Storage[]
          */
         private static $_instances = array();
+        /**
+         * @var bool|int
+         */
+        public $install_timestamp;
```
swashata, line 32: @DanieleAlessandra would it also work if you just add a `@property bool|int $install_timestamp` in the PHP doc block, instead of this change?

The reason I am asking is, let's not make changes to the existing system, unless needed.

includes/managers/class-fs-debug-manager.php
swashata, line 1: Great 👏

require.php
swashata, line 1: Revert indentation changes please.

swashata (changes requested):
Greatly done @DanieleAlessandra 👍 Please see my two comments and we can merge this.

Also don't forget to rebase on `develop` and bump pre-release version of `start.php`

includes/class-fs-storage.php
DanieleAlessandra: Maybe it will work, I try...

includes/class-freemius.php
```diff
@@ -20815,7 +20448,7 @@ private function _fetch_newer_version( $plugin_id = false, $flush = true, $expir
          *
          * @return bool|FS_Plugin_Tag
          */
-        function get_update( $plugin_id = false, $flush = true, $expiration = FS_Plugin_Updater::UPDATES_CHECK_CACHE_EXPIRATION, $newer_than = false ) {
+        function get_update( $plugin_id = false, $flush = true, $expiration = WP_FS__TIME_24_HOURS_IN_SEC, $newer_than = false ) {
```
swashata, line 20451: @DanieleAlessandra rebase glitch.. Please use `$expiration = FS_Plugin_Updater::UPDATES_CHECK_CACHE_EXPIRATION`

includes/class-freemius.php
```diff
@@ -21347,9 +20980,7 @@ private function _sync_plugin_license(
                 /**
                  * Sync licenses. Pass the site's license ID so that the foreign licenses will be fetched if the license
                  * associated with that ID is not included in the user's licenses collection.
-                 * Save previous value to manage remote license renewals.
                  */
-                $was_license_expired_before_sync = $this->_license->is_expired();
```
swashata, line 21352: @DanieleAlessandra rebase glitch, please revert

includes/class-freemius.php
```diff
@@ -22484,7 +22115,7 @@ private function check_updates(
             $background = false,
             $plugin_id = false,
             $flush = true,
-            $expiration = FS_Plugin_Updater::UPDATES_CHECK_CACHE_EXPIRATION,
+            $expiration = WP_FS__TIME_24_HOURS_IN_SEC,
```
swashata, line 22118: @DanieleAlessandra rebase glitch

swashata (changes requested):

swashata (changes requested):
Hey @DanieleAlessandra, nicely done. But it looks like I (accidentally) stumbled upon a race-condition or bad-state like issue in the class-fs-logger that can bite us.

Please see this

image

To reproduce, I have

* Changed `wp_schedule_single_event( time() + 24 * HOUR_IN_SECONDS, 'fs_debug_turn_off_logging_hook' );` to `wp_schedule_single_event( time() + 5, 'fs_debug_turn_off_logging_hook' );`
* Turned on `WP_DEBUG`
* Enabled the Freemius debugging.
* Waited 5 sec and refreshed
* The error pops up sometimes and sometimes it doesn't.

I think the issue is in class-fs-logger and the static variable `$_isStorageLoggingOn` which is not cleared/updated after calling `_set_storage_logging`. Kindly take a look. Thanks.

includes/class-fs-logger.php
```diff
@@ -11,6 +11,10 @@
 	}
 
 	class FS_Logger {
+        /**
+         * @var bool
+         */
+        public static  $ON          = false;
```
swashata, line 17: @DanieleAlessandra I think we are mixing two things here :)

We have the function `_set_storage_logging` which only deals with DB related storage. So making it `off` shouldn't really turn off regular logging.

## For calling `$wpdb->insert` when we have just turned off the logging

We need to update the already existing `$_isStorageLoggingOn` static variable inside the `_set_storage_logging`. Something like

```php
public static function _set_storage_logging( $is_on = true ) {
    // .. existing code

    self::$_isStorageLoggingOn = $is_on;
}
```

and that's it. I think this will work just fine.

## Out of sync DB creation issue

I see you've made adjustment in the creation statement with `CREATE TABLE IF NOT EXISTS`. I propose we do it like this:

1. Drop table if exists.
2. Then create the table.

I think this way when the logging is turned on we are 100% sure to get the table schema we are after. Since we are dropping table anyway when turning off, I think this approach is fine.

---

So kindly see if the above suggestion would work.

swashata (changes requested):

includes/class-fs-logger.php
```diff
@@ -125,6 +128,14 @@ function on() {
 			self::hook_footer();
 		}
 
+        /**
```
swashata, line 131: Indentation glitch

includes/class-fs-logger.php
```diff
@@ -240,6 +251,9 @@ function api_error( $api_result, $wrapper = false ) {
 		}
 
 		function entrance( $message = '', $wrapper = false ) {
+            if ( ! $this->is_on() ) {
```
swashata, line 254: Indentation glitch

swashata, line 254: This might not be needed since the `_log` is already bailing. If you want to be more memory/compute optimistic, then we need to have this check for every public log related function, but that isn't needed IMO.

includes/class-fs-logger.php
```diff
@@ -320,6 +334,11 @@ public static function _set_storage_logging( $is_on = true ) {
 
 			$table = "{$wpdb->prefix}fs_logger";
 
+            /**
```
swashata, line 337: Indentation glitch, also please fix for the rest of the file.

includes/class-fs-logger.php
```diff
@@ -329,7 +348,7 @@ public static function _set_storage_logging( $is_on = true ) {
 				 *
 				 * @link https://core.trac.wordpress.org/ticket/2695
 				 */
-				$result = $wpdb->query( "CREATE TABLE {$table} (
+				$result = $wpdb->query( "CREATE TABLE IF NOT EXISTS {$table} (
```
swashata, line 351: Is this `IF NOT EXISTS` needed now that we are dropping the table anyway?

includes/class-fs-logger.php
```diff
@@ -349,14 +368,21 @@ public static function _set_storage_logging( $is_on = true ) {
...
 KEY `type` (`type` ASC))" );
 			} else {
-				/**
-				 * Drop logging table.
-				 */
-				$result = $wpdb->query( "DROP TABLE IF EXISTS $table;" );
+                /**
+                 * Since logging table does not exist anymore, we need to turn off all instances
+                 */
+                foreach ( self::$LOGGERS as $logger ) {
```
swashata, line 374: I also think this isn't needed. Because we have just turned off "DB" storage of the log. The actual log happens with the constant `WP_FS__DEBUG_SDK`. So if DB storage is turned off, let there still be memory logging within the `_log` and it will show up in the JavaScript console. The only fix needed is turning off the `$_isStorageLoggingOn`  in line 385 like you did :)

includes/class-fs-logger.php
```diff
@@ -125,6 +128,14 @@ function on() {
 			self::hook_footer();
 		}
 
+        /**
+         * Turn off instance
+         *
+         * @return void
+         */
+        public function off() {
```
swashata, line 136: This method won't be needed anymore (please see my suggestion below). Also just a code convention thingy, we add a blank line after a function/method declaration.

swashata (changes requested):
Please see the indentation glitch and one conceptual thing @DanieleAlessandra

includes/managers/class-fs-debug-manager.php
```diff
@@ -0,0 +1,511 @@
...
+
+            if ( fs_request_is_post() && in_array( $is_on, array( 0, 1 ) ) ) {
+                if ( $is_on ) {
+                    self::_turn_on_debug_mode();
+                } else {
+                    self::_turn_off_debug_mode();
+                }
+
+                // Turn on/off storage logging.
+                FS_Logger::_set_storage_logging( ( 1 == $is_on ) );
```
swashata, line 291: @DanieleAlessandra please see that you are already setting storage login from the `_turn_off_debug_mode` or the `_turn_on_debug_mode` above (line 284). So I feel this call is redundant.

includes/class-fs-logger.php
DanieleAlessandra: Table could exist if it was created with a previous version and never deleted

swashata (approved):
Great job @DanieleAlessandra 👏

PR #692, opened 2024-03-14
[garbage-collector] Fixed an issue related to modified slugs.

fajardoleo:


includes/class-fs-garbage-collector.php
```diff
@@ -85,10 +87,30 @@ function clean() {
...
                         unset( $option[ "{$slug}:{$this->_type}" ] );
                         $updated = true;
-                    } else if ( isset( $product->id ) && array_key_exists( $product->id, $option ) ) { /* all_licenses */
-                        unset( $option[ $product->id ] );
-                        $updated = true;
-                    } else if ( isset( $product->file ) && array_key_exists( $product->file, $option ) ) { /* file_slug_map */
+                    } else if ( isset( $product->id ) && array_key_exists( $product->id, $option ) ) { /* all_licenses, add-ons, and id_slug_type_path_map */
+                        $unset_option = false;
+
+                        if ( ! isset( $active_products_by_id_map[ $product->id ] ) ) {
```
swashata, line 93: I think we can improve the readability here a little.

```php
$is_inactive_by_id = ! isset( $active_products_by_id_map[ $product->id ] );
$is_inactive_by_slug = (
    'id_slug_type_path_map' === $option_name &&
    (
        ! isset( $option[ $product->id ]['slug'] ) ||
        $slug === $option[ $product->id ]['slug']
    )
);

if ( $is_inactive_by_id || $is_inactive_by_slug ) {
    // do your thing 🥷 
}
```

swashata (changes requested):

swashata (approved):

Tobias-Conrad:
@fajardoleo @swashata i and one client affected. So we did the fixer plugin and good.

I feel that change is freemius internal, just a fix of the collector and should be released ASAP as fixes should be released.


As the freemius fixer is often a solution, build it inside the sdk and it would solve this problem and the others with licenses.
Add a reset license that trigger the fixer...

PR #697, opened 2024-03-25
[license] [sticky] [renewal] Remove sticky message on extended license

DanieleAlessandra:
Fixes a bug where, if a license expires on a WordPress site and is subsequently renewed on the server side, the license expiration notice persists on the WordPress site even after the license details are synchronized with our server.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -21347,7 +21347,9 @@
                 /**
                  * Sync licenses. Pass the site's license ID so that the foreign licenses will be fetched if the license
                  * associated with that ID is not included in the user's licenses collection.
+                 * Save previous value to manage remote license renewals.
                  */
+                $was_license_expired_before_sync = $this->_license->is_expired();
                 $this->_sync_licenses(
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -21473,6 +21475,14 @@ private function _sync_plugin_license(
                                     $plan_change = 'expired';
                                 }
                             }
+						} else if ( $was_license_expired_before_sync ) {
+							/**
+							 * If license was expired but it is not anymore
```
swashata, line 21480: Please add a period `.` after the sentence.

swashata (changes requested):
Hey @DanieleAlessandra The changes look good. Please fix the indentation (we are using spaces and not tabs) and then we can merge your PR.

Also kindly see that in `start.php`, the `$this_sdk_version = '2.6.2.1';` variable. Please bump it to the next patch `$this_sdk_version = '2.6.2.2';`. We follow this while merging something to the develop, so that people can test it out.

PR #699, opened 2024-03-28
[updates-check] [api] [cache] Improved the updates check API response…

fajardoleo:
… caching.

### DIFF
--- a/includes/class-fs-api.php
+++ b/includes/class-fs-api.php
@@ -318,9 +318,12 @@
 				$flush = true;
 			}
 
-			$cached_result = self::$_cache->get( $cache_key );
+			$has_valid_cache = self::$_cache->has_valid( $cache_key, $expiration );
+			$cached_result   = $has_valid_cache ?
+				self::$_cache->get( $cache_key ) :
+				null;
 
-			if ( $flush || ! self::$_cache->has_valid( $cache_key, $expiration ) ) {
+			if ( $flush || is_null( $cached_result ) ) {
 				$result = $this->call( $path );
 
 				if ( ! is_object( $result ) || isset( $result->error ) ) {
--- a/includes/class-fs-plugin-updater.php
+++ b/includes/class-fs-plugin-updater.php
@@ -37,6 +37,8 @@
 
         private static $_upgrade_basename = null;
 
+        const UPDATES_CHECK_CACHE_EXPIRATION = ( WP_FS__TIME_24_HOURS_IN_SEC / 24 );
+
         #--------------------------------------------------------------------------------
         #region Singleton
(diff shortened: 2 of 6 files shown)


includes/class-fs-api.php
```diff
@@ -318,9 +318,12 @@ function get( $path = '/', $flush = false, $expiration = WP_FS__TIME_24_HOURS_IN
 				$flush = true;
 			}
 
-			$cached_result = self::$_cache->get( $cache_key );
+            $has_valid_cache = self::$_cache->has_valid( $cache_key, $expiration );
```
swashata, line 321: @fajardoleo please use `tab` for indentation. The rest of the file is using tab.

includes/class-fs-api.php
```diff
@@ -318,9 +318,12 @@ function get( $path = '/', $flush = false, $expiration = WP_FS__TIME_24_HOURS_IN
...
 			}
 
-			$cached_result = self::$_cache->get( $cache_key );
+            $has_valid_cache = self::$_cache->has_valid( $cache_key, $expiration );
+            $cached_result   = $has_valid_cache ?
+                self::$_cache->get( $cache_key ) :
+                null;
 
-			if ( $flush || ! self::$_cache->has_valid( $cache_key, $expiration ) ) {
+            if ( $flush || ! $has_valid_cache ) {
```
swashata, line 326: I suggest to do it like

```php
if ( $flush || is_null( $cached_result ) {
```

This is just for self-guarding code, where the `$_cache->get` would return `null` if not present.

includes/class-fs-plugin-updater.php
```diff
@@ -1188,7 +1190,7 @@ private static function get_tested_wp_version( $tested_up_to ) {
          * @return object
          */
         private function get_latest_download_details( $addon_id = false, $newer_than = false, $fetch_readme = true ) {
-            return $this->_fs->_fetch_latest_version( $addon_id, true, WP_FS__TIME_24_HOURS_IN_SEC, $newer_than, $fetch_readme );
+            return $this->_fs->_fetch_latest_version( $addon_id, true, FS_Plugin_Updater::UPDATES_CHECK_CACHE_EXPIRATION, $newer_than, $fetch_readme );
```
swashata, line 1193: Only here it seems the cache time is changed from 24hrs to 1hr. Please let me know if this is intentional.

swashata (changes requested):
Hi @fajardoleo , I did my best reviewing it. Please see my comment. We can also sync on slack. Also kindly rebase on `develop`. Thanks.

swashata (approved):

PR #702, opened 2024-04-09
[i18n] Update the strings for case sensitivity. Remove lines spacings in translatable strings.

bahiirwa:
- [x] Update the "Cancel" strings for case sensitivity.

### DIFF
--- a/languages/freemius.pot
+++ b/languages/freemius.pot
@@ -8,7 +8,7 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Language-Team: Freemius Team <admin@freemius.com>\n"
 "Last-Translator: Vova Feldman <vova@freemius.com>\n"
-"POT-Creation-Date: 2024-04-21 11:18+0000\n"
+"POT-Creation-Date: 2024-04-21 11:55+0000\n"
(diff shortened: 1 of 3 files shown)


includes/class-freemius.php
```diff
@@ -4827,9 +4827,9 @@ function dynamic_init( array $plugin_info ) {
                         /* translators: %s: License type (e.g. you have a professional license) */
                             $this->get_text_inline( 'You have purchased a %s license.', 'you-have-x-license' ),
                             fs_request_get( 'purchased_plan' )
-                        ) .
+                        ) . ' ' .
                         sprintf(
-                            $this->get_text_inline(" The %s's %sdownload link%s, license key, and installation instructions have been sent to %s. If you can't find the email after 5 min, please check your spam box.", 'post-purchase-email-sent-message' ),
```
swashata, line 4832: @bahiirwa please revert this and other changes with the whitespace :) I know it is not ideal, but we will lose all community translations over this. Kindly just focus on fixing the `cancel` button text. Thanks.

swashata (changes requested):

swashata (approved):
LGTM

PR #703, opened 2024-04-10
[cleanup] Remove unused mixins from the codebase

bahiirwa:
- [x] Remove unused mixins from the codebase.
- [x] Use the autoprefixer in the build process. Remove mixins made obsolute. We have an autoprefixer which makes the build process more efficient in managing future browser support and deprecations.

### DIFF
--- a/assets/scss/_mixins.scss
+++ b/assets/scss/_mixins.scss
@@ -22,15 +22,6 @@
 //  THE SOFTWARE.
 //
 
-// ---- LEGACY IE SUPPORT USING FILTERS ----
-//  Should IE filters be used or not?
-//  PROS: gradients, drop shadows etc will be handled by css.
-//  CONS: will harm the site performance badly, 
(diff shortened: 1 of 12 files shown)


DanieleAlessandra (approved):
I have reviewed the modifications to the SCSS files and observed that the resulting CSS is essentially unchanged from before. Initially, I had reservations about the continued necessity of maintaining vendor prefixes in 2024. However, I believe that we should continue to support them for a while. Therefore, I am satisfied with the current state of the changes.

swashata:
> I have reviewed the modifications to the SCSS files and observed that the resulting CSS is essentially unchanged from before. Initially, I had reservations about the continued necessity of maintaining vendor prefixes in 2024. However, I believe that we should continue to support them for a while. Therefore, I am satisfied with the current state of the changes.

Thanks for the review @DanieleAlessandra, we already have POSTCSS which adds the prefix if needed.

PR #705, opened 2024-04-19
[privacy] [activation] If there is no user during activation shows a …

DanieleAlessandra:
Do not send user information on license activation. If API responds that it was not able to find a user linked to license, it shows a small form to gather requested informations.


includes/class-freemius.php
```diff
@@ -17214,9 +17223,6 @@ function get_opt_in_params( $override_with = array(), $network_level_or_blog_id
             $versions = $this->get_versions();
 
             $params = array_merge( $versions, array(
-                'user_firstname'    => $current_user->user_firstname,
-                'user_lastname'     => $current_user->user_lastname,
-                'user_email'        => $current_user->user_email,
```
fajardoleo, line 17219: Since the `$current_user` variable is no longer used, we can remove the declaration above.

templates/connect.php
```diff
@@ -324,7 +324,34 @@ class="wrap<?php if ( ! fs_is_network_admin() && ( ! $fs->is_enable_anonymous()
...
+                        The fields below are used to explicitly collect missing information in cases where it is not possible
+                        to assign an existing user to the license being activated. This was done to avoid sending private data
+                         in any case.
+                    -->
+                    <span class="fs-message"><?php fs_echo_inline( "Please enter at least a valid email address to activate your license:",
+                            'email-is-requested' ) ?></span>
+                    <div class="fs-fields-container">
+                        <div class="fs-input-container">
+                            <label for="first_name" class="screen-reader-text">First name</label>
+                            <input type="text" id="first_name" name="first_name" placeholder="First name"
```
fajardoleo, line 339: Missing the `fs_` prefix for the `id` attribute.

templates/connect.php
```diff
@@ -525,6 +552,11 @@ class="button button-secondary" tabindex="2"><?php fs_esc_html_echo_x_inline( 'S
 				$( document.body ).addClass( 'fs-loading' );
 			};
 
+            $('#fs_personal_data input').on('focus', function() {
+                $('#fs_personal_data').removeClass('error');
+                resetLoadingMode();
+            })
```
fajardoleo, line 558: Missing a semi-colon (https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#semicolons).

templates/connect.php
```diff
@@ -525,6 +552,11 @@ class="button button-secondary" tabindex="2"><?php fs_esc_html_echo_x_inline( 'S
 				$( document.body ).addClass( 'fs-loading' );
 			};
 
+            $('#fs_personal_data input').on('focus', function() {
```
fajardoleo, line 555: We also need spaces here (similar to PHP). There are examples here:
https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#examples-of-good-spacing

So it should be like this:
```
$( '#fs_personal_data input' ).on( 'focus', function() {
    $( '#fs_personal_data' ).removeClass( 'error' );
    resetLoadingMode();
} );
```

templates/connect.php
```diff
@@ -697,6 +729,19 @@ function updatePrimaryCtaText( actionType ) {
 		var ajaxOptin = ( requireLicenseKey || isNetworkActive );
 
 		$form.on('submit', function () {
+            const personal_data = {};
+            if ($personalData.is(':visible')) {
```
fajardoleo, line 733: Please check the comment above about spacing.

fajardoleo (changes requested):

PR #715, opened 2024-05-22
[opt-in] Fixed an issue with the opt-in moderation and API connectivi…

fajardoleo:
…ty test logic.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -4174,12 +4174,16 @@
                 ! empty( $this->_storage->connectivity_test ) &&
                 isset( $this->_storage->connectivity_test['is_active'] )
             ) {
-                $is_active = $this->_storage->connectivity_test['is_active'];
+                $is_connected = isset( $this->_storage->connectivity_test['is_connected'] ) ?
+                    $this->_storage->connectivity_test['is_connected'] :
+                    null;
+                $is_active    = ( $this->_storage->connectivity_test['is_active'] || is_object( $this->_site ) );
             } else {
-                $is_active = $this->should_turn_fs_on( $this->apply_filters( 'is_plugin_update', $this->is_plugin_update() ) );
-
-                $this->store_connectivity_info( (object) array( 'is_active' => $is_active ), null );
+                $is_connected = null;
+                $is_active    = $this->should_turn_fs_on( $this->apply_filters( 'is_plugin_update', $this->is_plugin_update() ) );
             }
+
+            $this->store_connectivity_info( (object) array( 'is_active' => $is_active ), $is_connected );
 
             if ( $is_active ) {
                 $this->_is_on = true;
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -4174,13 +4174,17 @@ function has_api_connectivity( $flush_if_no_connectivity = false ) {
...
+                $is_active    = ( $this->_storage->connectivity_test['is_active'] || is_object( $this->_site ) );
             } else {
-                $is_active = $this->should_turn_fs_on( $this->apply_filters( 'is_plugin_update', $this->is_plugin_update() ) );
-
-                $this->store_connectivity_info( (object) array( 'is_active' => $is_active ), null );
+                $is_connected = null;
+                $is_active    = $this->should_turn_fs_on( $this->apply_filters( 'is_plugin_update', $this->is_plugin_update() ) );
             }
 
+            $this->store_connectivity_info( (object) array( 'is_active' => $is_active ), $is_connected );
```
swashata, line 4186: @fajardoleo A question, would it work if we do it like this instead

```php
if ( $is_active ) {
  $this->_is_on = true;
  $this->update_connectivity_info( $is_connected );
}
```

Maybe it will also save us to not cache the connectivity in cache in case it could not be determined?

includes/class-freemius.php
```diff
@@ -5484,7 +5488,11 @@ private function parse_settings( &$plugin_info ) {
                 'affiliate_moderation' => $this->get_option( $plugin_info, 'has_affiliation' ),
                 'bundle_id'            => $this->get_option( $plugin_info, 'bundle_id', null ),
                 'bundle_public_key'    => $this->get_option( $plugin_info, 'bundle_public_key', null ),
-                'opt_in_moderation'    => $this->get_option( $plugin_info, 'opt_in', null ),
+                'opt_in_moderation'    => $this->get_option(
+                    $plugin_info,
+                    'opt_in',
+                    $this->get_option( $plugin_info, 'opt_in_moderation', null )
```
swashata, line 5494: Perhaps just add a comment here that we support both parameter names `opt_in` and `opt_in_moderation` for backward compatibility.

swashata (commented):
Please see my comment, thanks.

includes/class-freemius.php
fajardoleo: @swashata Looks great! By the way, since `$this->store_connectivity_info()` updates `$this->_has_api_connection`, it will look like this:
```
if ( $is_active ) {
    $this->_is_on = true;
    $this->update_connectivity_info( $is_connected );
} else if ( $is_connected ) {
    $this->_has_api_connection = true;
}

return $this->_has_api_connection;
```

But please check to see if there's an alternative to the `else if` part. Thanks!

swashata (approved):
LGTM

PR #718, opened 2024-06-26
[plan-entity] Added missing FS_Plugin_Plan property declarations to p…

fajardoleo:
…revent triggering warnings when the plugin information dialog is opened.


swashata (approved):
LGTM

PR #722, opened 2024-07-17
[filters] [pricing-url] Updated the pricing_url filter so that it als…

fajardoleo:
…o updates the pricing/upgrade submenu item's URL.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -14674,6 +14674,20 @@
                 $this->_parent->addon_url( $this->_slug ) :
                 $this->_get_admin_page_url( 'pricing', $params );
 
+            return $this->get_pricing_url_with_filter( $url );
+        }
+
+        /**
+         * Retrieves the filtered pricing URL.
+         *
+         * @author Leo Fajardo (@leorw)
+         * @since  2.7.4
+         *
+         * @param string $url
+         *
+         * @return string
+         */
+        private function get_pricing_url_with_filter( $url ) {
             return $this->apply_filters( 'pricing_url', $url );
         }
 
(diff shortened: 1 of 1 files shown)


includes/class-freemius.php
```diff
@@ -18800,18 +18800,34 @@ private function add_submenu_items() {
...
-                        array( &$this, '_pricing_page_render' ),
-                        $this->get_plugin_name() . ' &ndash; ' . $this->get_text_x_inline( 'Pricing', 'noun', 'pricing' ),
-                        'manage_options',
-                        'pricing',
-                        'Freemius::_clean_admin_content_section',
-                        WP_FS__LOWEST_PRIORITY,
-                        ( $add_submenu_items && $show_pricing ),
-                        $pricing_class
-                    );
+                    $custom_pricing_url        = $this->apply_filters( 'pricing_url', null );
```
swashata, line 18803: @fajardoleo Could you please refactor the `pricing_url` method and extract the logic to apply the filter in a private function

```php
function pricing_url(...) {
  $url = ...;

  return $this->get_pricing_url_with_filter( $url );
}

private function get_pricing_url_with_filter( $url ) {
  return $this->apply_filters( 'pricing_url', $url );
}
```

Then here, instead of calling the filter ad-hoc, you can do

```php
$custom_pricing_url = $this->get_pricing_url_with_filter( null );
```

This will result in a single source of truth where the filter `pricing_url` is applied so it less error prone. Thanks.

swashata (changes requested):

swashata (approved):
LGTM 👏

PR #724, opened 2024-07-17
[plugin-information-dialog] Try to fetch data from WP only when the p…

fajardoleo:
…lugin is actually WP.org-compliant.


swashata (approved):
LGTM

PR #726, opened 2024-07-18
[plugin-information-dialog] [updates] Updated the update mechanism to…

fajardoleo:
… align with the changes in WordPress plugin update AJAX requests.


swashata (approved):
LGTM, also tested with both both WP 6.5, 6.6 and below.

PR #727, opened 2024-07-23
Improved the support for plugins with free and premium versions that …

fajardoleo:
…can be activated in parallel.


includes/class-freemius.php
```diff
@@ -7407,6 +7407,11 @@ function _activate_plugin_event_hook() {
                     $this->apply_filters( 'deactivate_on_activation', true )
                 ) {
                     deactivate_plugins( $other_version_basename );
+                } else if ( $is_premium_version_activation ) {
```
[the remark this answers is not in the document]
swashata: Now thinking about it again, perhaps we would want to also handle "Free plugin activation", for example user activates pro version, then deactivates free version on their own. Then notices the pro doesn't work without the Free, so they activate the Free version again. So instead of the `else` part here, we can do something like (pseudo code)

```php
// If both free and premium version are active, then remove SDK from the Free version to avoid having it load the configuration.
if (is_plugin_active( $other_version_basename ) ) {
  fs_remove_sdk_reference_by_basename( $this->_free_plugin_basename );
}
```

But please double check me.

swashata (commented):
Please see my suggestions, thanks.

includes/class-freemius.php
```diff
@@ -8198,20 +8203,35 @@ function _deactivate_plugin_hook() {
...
                 }
             }
 
-            fs_fallback_to_newest_active_sdk();
+            if (
+                ! empty( $fs_active_plugins->newest ) &&
+                ! empty( $fs_active_plugins->newest->plugin_path ) &&
+                $fs_active_plugins->newest->plugin_path === $plugin_basename
+            ) {
+                fs_fallback_to_newest_active_sdk();
```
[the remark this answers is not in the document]
fajardoleo: @swashata It was actually supposed to reduce risks since it's called only when it's necessary. However, in the current code, we are always calling it anyway. I have reverted the changes.

includes/class-freemius.php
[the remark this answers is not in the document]
fajardoleo: @swashata In that use case, when the premium version is activated, the latest SDK is already set to the SDK that comes with the premium version (when using the code in this PR). When the free version is deactivated and reactivated, it will not have any effect on the latest SDK information. Please let me know if I missed something :) Thanks!

includes/fs-essential-functions.php
```diff
@@ -415,4 +415,25 @@ function fs_fallback_to_newest_active_sdk() {
                 fs_update_sdk_newest_version( $newest_sdk_path, $newest_sdk_data->plugin_path );
             }
         }
+    }
+
+    if ( ! function_exists( 'fs_remove_sdk_reference_by_basename' ) ) {
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.7.4
```
swashata, line 423: `@since 2.8.0` or `2.7.3.6`. I don't think there will be a `2.7.0` version.

includes/class-freemius.php
```diff
@@ -7544,6 +7545,45 @@ function _activate_plugin_event_hook() {
...
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.8.0
+         *
+         * @param string $plugin
+         * @param bool   $network_wide
+         */
+        function _activated_plugin( $plugin, $network_wide ) {
+            // We cannot rely on fs_remove_sdk_reference_by_basename() for reordering, as the latest SDK reference might be associated with a different product, leading to the reordering of that product instead of the intended one.
+            $this->move_free_plugin_to_end_of_active_plugins( $this->_free_plugin_basename );
```
swashata, line 7557: I suggest not to pass the `$this->_free_plugin_basename` as the function can already get it from the instance. (Just for code readability)

includes/class-freemius.php
```diff
@@ -7544,6 +7545,45 @@ function _activate_plugin_event_hook() {
...
+
+            if ( $plugin !== $this->_free_plugin_basename ) {
+                /**
+                 * To avoid placing the free version at the start of the active plugins option when the SDK is loaded from it, remove the SDK reference linked to the free version if it is the newest.
+                 */
+                fs_remove_sdk_reference_by_basename( $this->_free_plugin_basename );
```
swashata, line 7563: Will it hurt us even if the SDK reference from the free version is at top? I don't mind doing this "just in case", but just asking for my understanding.

includes/class-freemius.php
```diff
@@ -7544,6 +7545,45 @@ function _activate_plugin_event_hook() {
...
+            }
+        }
+
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.8.0
+         *
+         * @param string $plugin
+         */
+        private function move_free_plugin_to_end_of_active_plugins( $plugin ) {
```
swashata, line 7573: Please remove the `$plugin` parameter and rely solely on `$this->_free_plugin_basename`. If that's not set, just bail early.

```php
if ( empty( $this->_free_plugin_basename ) ) {
  return;
}
```

includes/class-freemius.php
```diff
@@ -7544,6 +7545,45 @@ function _activate_plugin_event_hook() {
...
+            $active_plugins = get_option( 'active_plugins' );
+
+            // Move the free version to the last position in the active plugins option.
+            if ( false !== ( $key = array_search( $plugin, $active_plugins ) ) ) {
+                unset( $active_plugins[ $key ] );
```
swashata, line 7578: Let's either use `array_splice` or right after `unset`, let's reindex the array with `$active_plugins = array_values( $active_plugins )`.

includes/class-freemius.php
swashata, line 7573: Don't we want to handle network wide activation too?

CleanShot 2024-08-20 at 11 35 31@2x

Maybe it will be difficult to support network wide. Since I see the `wp_get_active_network_plugins` function does a `sort` when being called. If that's the case then please add some comment here.

includes/class-freemius.php
```diff
@@ -7544,6 +7545,45 @@ function _activate_plugin_event_hook() {
...
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since 2.8.0
+         *
+         * @param string $plugin
+         * @param bool   $network_wide
+         */
+        function _activated_plugin( $plugin, $network_wide ) {
```
swashata, line 7555: Proposing to rename this function to `_handle_free_and_premium_parallel_activation` or something like that, just for readability.

swashata (changes requested):

PR #728, opened 2024-07-29
[trial] [dialog] Fixes trial activation button layout

DanieleAlessandra:
Added a `display:flex` container to keep text and button separate and consistent in different screen sizes


fajardoleo (approved):
Looks good.

PR #730, opened 2024-07-31
[license-dialog] Fixed the retrieving of the selected license's ID fr…

fajardoleo:
…om the license update dialog.


fajardoleo:
Closes #747

PR #734, opened 2024-08-07
Feature/daniele fix huge cache

DanieleAlessandra:
In FS_Api::get added a check to avoid caching the result og `latest.json` because it usually contains all README.txt content and could be too long.


fajardoleo (approved):
Looks good to me.

PR #737, opened 2024-08-14
Fix bug with notice bar covering menu

DanieleAlessandra:
Set z-index of `.fs-secure-notice` to 9989 in order to avoid covering `#adminmenuwrap` that has z-index: 9990


swashata (approved):
LGTM

PR #738, opened 2024-08-19
[POC] Feature/swas/try parallel activation

swashata:
…d paid plugins.


swashata:
Not feasible as it can break the updater and other things such as code type changes.

PR #743, opened 2024-09-24
[privacy] [activation] Manage response with no user from request nor from license

DanieleAlessandra:


DanieleAlessandra:
This PR is wrong, we should target `develop` branch

PR #746, opened 2024-10-02
Feature/daniele/enrich debug page

DanieleAlessandra:
Enriched Debug Page:

- Added: unfiltered_site_url;
- Added: Scheduled Crons
- Added: Collapsile Tables
- Fixed: CSV Logs Download


includes/class-freemius.php
```diff
@@ -22006,6 +22006,9 @@ function _fetch_latest_version(
 
             if ( true === $fetch_readme ) {
                 $latest_version_endpoint = add_query_arg( 'readme', 'true', $latest_version_endpoint );
+
+                // Don't cache the API response when fetching readme information.
+                $expiration = null;
```
swashata, line 22011: How come this code is here @DanieleAlessandra ? Please have a look. You are targeting `master` instead of `develop` ;)

swashata (changes requested):
I will do proper review once you've changed the base to `develop` and made sure there aren't any commit which shouldn't be here.

PR #748, opened 2024-10-11
Feature/daniele/enrich debug page

DanieleAlessandra:
Enriched Debug Page:

- Added: unfiltered_site_url;
- Added: Scheduled Crons
- Added: Collapsile Tables
- Fixed: CSV Logs Download


templates/debug.php
```diff
@@ -114,7 +114,9 @@ function stopCountdownManually() {
         }(jQuery));
     </script>
 </div>
-<h2><?php fs_esc_html_echo_inline( 'Actions', 'actions' ) ?></h2>
+<h2>
+    <?php fs_esc_html_echo_inline( 'Actions', 'actions' ) ?>
+</h2>
```
fajardoleo, line 119: @DanieleAlessandra I suggest reverting this as the text remains the same.

templates/debug/scheduled-crons.php
```diff
@@ -73,7 +84,18 @@
 
 	$sec_text = fs_text_x_inline( 'sec', 'seconds' );
 ?>
-<h1><?php fs_esc_html_echo_inline( 'Scheduled Crons' ) ?></h1>
+<?php echo $title_tag_open; ?>
+<?php
+	if ($display_toggle_button) {
```
fajardoleo, line 89: @DanieleAlessandra There are missing spaces here.

fajardoleo, line 89: @DanieleAlessandra How about this instead?

```
<?php if ( $display_toggle_button ) : ?>
<button class="toggle-table" aria-expanded="<?php echo $is_open ?>">
    <span class="caret-icon">▼</span>
</button>
<?php endif ?>
```

Also, in our convention, we don't add a semicolon for single-line PHP code.

templates/debug/scheduled-crons.php
```diff
@@ -13,6 +13,17 @@
 	$fs_options      = FS_Options::instance( WP_FS__ACCOUNTS_OPTION_NAME, true );
 	$scheduled_crons = array();
 
+	$title_tag             = isset($VARS['title_tag']) ? $VARS['title_tag'] : 'h1';
```
fajardoleo, line 16: @DanieleAlessandra Missing spaces. I recommend reviewing this and adding the relevant spacing accordingly:
https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#whitespace

templates/debug/scheduled-crons.php
```diff
@@ -13,6 +13,17 @@
...
+	$title_tag_open        = "<$title_tag>";
+	$title_tag_close       = "</$title_tag>";
+	$display_toggle_button = isset($VARS['display_toggle']) ? $VARS['display_toggle'] : false;
+	if (isset($VARS['is_open'])) {
+		$is_open = $VARS['is_open'] ? 'true' : 'false';
+	}
+	else {
```
fajardoleo, line 23: @DanieleAlessandra I recommend reviewing this for brace style:
https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#brace-style

templates/debug.php
```diff
@@ -714,9 +750,19 @@ function stopCountdownManually() {
         </table>
     <?php endif ?>
 <?php endforeach ?>
+<?php
+    $page_params = array(
+        'title_tag' => 'h2',
+        'display_toggle' => true,
+        'is_open' => false,
+    );
+    fs_require_template( 'debug/scheduled-crons.php', $page_params );
```
fajardoleo, line 759: @DanieleAlessandra We can have some alignment like this:
```
$page_params = array(
    'title_tag'            => 'h2',
    'display_toggle' => true,
    'is_open'            => false,
);

fs_require_template( 'debug/scheduled-crons.php', $page_params );
```

Also, I added an empty line above `fs_require_template()`.

includes/class-fs-logger.php
```diff
@@ -691,5 +695,38 @@ public static function get_logs_download_url( $filename = '' ) {
...
 		}
 
+		/**
+		 * @param $file_path
+		 * @param $query_results
+		 *
+		 * @return bool
+		 */
+		private static function write_csv_to_filesystem( $file_path, $query_results ) {
+			require_once( ABSPATH . 'wp-admin/includes/file.php' );
```
fajardoleo, line 705: @DanieleAlessandra I think it should be like this:
```
if ( empty( $query_results ) ) {
    return false;
}

require_once( ABSPATH . 'wp-admin/includes/file.php' );

WP_Filesystem();
global $wp_filesystem;

$headers = array_keys( (array) $query_results[0] );
```

This way, we won't trigger the other lines if we will return `false` anyway.

Also, the `require_once` line doesn't appear to be needed, as we are in a WordPress admin context where that is already included.

includes/class-fs-logger.php
```diff
@@ -691,5 +695,38 @@ public static function get_logs_download_url( $filename = '' ) {
 			return rtrim( $upload_dir['url'], '/' ) . $filename;
 		}
 
+		/**
+		 * @param $file_path
+		 * @param $query_results
```
fajardoleo, line 700: @DanieleAlessandra Parameter types are missing.

includes/class-fs-logger.php
```diff
@@ -691,5 +695,38 @@ public static function get_logs_download_url( $filename = '' ) {
...
+			WP_Filesystem();
+			global $wp_filesystem;
+
+			if ( ! empty( $query_results ) ) {
+				$headers = array_keys( (array) $query_results[0] );
+			} else {
+				return false;
+			}
+
+			$content = implode( "\t", $headers ) . "\n";
```
fajardoleo, line 715: @DanieleAlessandra The headers are duplicated by the `foreach()` loop below.

templates/debug.php
```diff
@@ -888,3 +934,32 @@ class="dashicons dashicons-download"></i> <?php fs_esc_html_echo_inline( 'Downlo
         });
     </script>
 <?php endif ?>
+<script type="text/javascript">
+    // JavaScript to toggle the visibility of the table body and change the caret icon
+    document.addEventListener('DOMContentLoaded', function () {
+        document.querySelectorAll('.toggle-table').forEach(function (button) {
```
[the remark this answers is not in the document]
DanieleAlessandra: While the codebase still contains some jQuery usage, I believe it is beneficial for us to reduce dependency on it over time, as it’s likely to be phased out in the future.

For this specific block, I opted not to use jQuery because it is entirely new and independent from the other scripts on the page. Additionally, the functionality is simple enough to implement without relying on the library.

templates/debug.php
```diff
@@ -888,3 +931,37 @@ class="dashicons dashicons-download"></i> <?php fs_esc_html_echo_inline( 'Downlo
...
+
+            button.on( 'click', function () {
+                toggle( button.attr( 'aria-expanded' ) === 'false' );
+            } );
+
+            toggle( button.attr( 'aria-expanded' ) === 'true' );
+        } );
+    } );
+</script>
+<style>
```
fajardoleo, line 959: @DanieleAlessandra It seems that the styles here should be in the `assets/sass/admin/debug.scss` file.

templates/debug.php
```diff
@@ -284,7 +288,12 @@ function stopCountdownManually() {
         <?php endforeach ?>
     </tbody>
 </table>
-<h2><?php fs_esc_html_echo_x_inline( 'SDK Versions', 'as software development kit versions', 'sdk-versions' ) ?></h2>
+<h2>
+    <button class="toggle-table" aria-expanded="true">
```
fajardoleo, line 292: @DanieleAlessandra I see that this is duplicated in multiple places, so I suggest storing it in a variable and then doing something like `<?php echo $toggle_button ?>`.

Also, the class should have the `fs-` prefix. It could be something like `fs-debug-table-toggle-button`. Is the `caret-icon` class needed?

templates/debug.php
```diff
@@ -888,3 +931,37 @@ class="dashicons dashicons-download"></i> <?php fs_esc_html_echo_inline( 'Downlo
...
+            const button = $( this );
+            const table = button.closest( 'h2' ).next( 'table' );
+            table.css( 'overflow', 'hidden' );
+
+            const toggle = function ( isExpanded ) {
+                button.attr( 'aria-expanded', isExpanded );
+                button.find( '.caret-icon' ).text( isExpanded ? '▼' : '▲' );
+                table.css( {
+                    display  : isExpanded ? 'table' : 'block',
+                    maxHeight: isExpanded ? 'auto' : '0'
```
fajardoleo, line 947: @DanieleAlessandra Maybe also add `borderBottomWidth: isExpanded ? '1px' : '0',`.

templates/debug.php
```diff
@@ -714,6 +748,15 @@ function stopCountdownManually() {
...
     <?php endif ?>
 <?php endforeach ?>
+<?php
+	$page_params = array(
+		'title_tag'      => 'h2',
+		'display_toggle' => true,
+		'is_open'        => false,
+	);
+
+	fs_require_template( 'debug/scheduled-crons.php', $page_params );
```
fajardoleo, line 758: @DanieleAlessandra Regarding the duplicated `<button class="toggle-table" aria-expanded="true">`, if it's stored in a variable, then it can also be included in `$page_params`.

templates/debug.php
DanieleAlessandra: Not relevant anymore because final array only has one element

fajardoleo (approved):
Looks good to me.

PR #749, opened 2024-10-13
[anonymous-mode] Support anonymous mode by default on sandbox sites.

fajardoleo:


DanieleAlessandra (approved):
Tested with local configuration and with TasteWP.

PR #752, opened 2024-10-22
Added a boolean parameter to get_plan_by_id to avoid calling sync_pla…

DanieleAlessandra:
…ns when coming from sync_plan


fajardoleo (approved):
Looks good to me.

DanieleAlessandra:
Closes #757

PR #756, opened 2024-11-05
[composer] [autoload] Added start.php to autoload.files in composer.json

DanieleAlessandra:


start.php
```diff
@@ -83,11 +83,25 @@ function_exists( 'wp_is_json_request' ) &&
      */
 	$themes_directory         = get_theme_root( get_stylesheet() );
 	$themes_directory_name    = basename( $themes_directory );
-	$theme_candidate_basename = basename( dirname( $fs_root_path ) ) . '/' . basename( $fs_root_path );
 
-	if ( $file_path == fs_normalize_path( realpath( trailingslashit( $themes_directory ) . $theme_candidate_basename . '/' . basename( $file_path ) ) )
-	) {
-		$this_sdk_relative_path = '../' . $themes_directory_name . '/' . $theme_candidate_basename;
+	// This change ensures that the condition works even if the SDK is located in a subdirectory (e.g., vendor)
+	$theme_candidate_basename = str_replace($themes_directory . '/' . get_stylesheet() . '/', '', $fs_root_path );
```
swashata, line 88: @DanieleAlessandra While I understand your logic, I think the placement of it can be made better.

The variable `$theme_candidate_basename` means the "basename" of the theme itself, it doesn't need to be the basename of the SDK directory inside the theme.

So either

1. Rename this variable to something like `$theme_candidate_sdk_basename`
2. Or change the logic of `$is_current_active_theme`

start.php
```diff
@@ -83,11 +83,25 @@ function_exists( 'wp_is_json_request' ) &&
...
+
+	// Check if the current file is part of the active theme.
+	$is_current_active_theme = $file_path == $themes_directory . '/' . get_stylesheet() . '/' . $theme_candidate_basename . '/' . basename($file_path);
```
swashata, line 91: I would name the variable like

1. `$is_current_active_theme` --> `$is_current_sdk_from_active_theme`
2. `$is_current_parent_theme` --> `$is_current_sdk_from_parent_theme`.

start.php
```diff
@@ -176,7 +190,8 @@ function_exists( 'wp_is_json_request' ) &&
 	     $this_sdk_version != $fs_active_plugins->plugins[ $this_sdk_relative_path ]->version
 	) {
 		if ( $is_theme ) {
-			$plugin_path = basename( dirname( $this_sdk_relative_path ) );
+			// Saving relative path and not only directory name as it could be a subfolder
+			$plugin_path = $this_sdk_relative_path;
```
swashata, line 194: This is creating an anomaly in logic. For plugins, we are saving the basename of the plugin, but for theme, we are saving the SDK's relative path! Can't we do better here? That is for both plugin and theme, save the basename?

start.php
```diff
@@ -229,7 +244,8 @@ function_exists( 'wp_is_json_request' ) &&
 			$is_newest_sdk_plugin_active = is_plugin_active( $fs_newest_sdk->plugin_path );
 		} else {
 			$current_theme               = wp_get_theme();
-			$is_newest_sdk_plugin_active = ( $current_theme->stylesheet === $fs_newest_sdk->plugin_path );
+			// Detect if current theme is the one registered as newer SDK
+			$is_newest_sdk_plugin_active = ( strpos( $fs_newest_sdk->plugin_path, '/' . $themes_directory_name . '/' . $current_theme->stylesheet . '/' ) !== false );
```
swashata, line 248: Please see my suggestion above, perhaps we can skip this change!

start.php
```diff
@@ -238,7 +254,8 @@ function_exists( 'wp_is_json_request' ) &&
              * from happening by keeping the SDK info stored in the `fs_active_plugins` option.
              */
             if ( ! $is_newest_sdk_plugin_active && $current_theme_parent instanceof WP_Theme ) {
-                $is_newest_sdk_plugin_active = ( $fs_newest_sdk->plugin_path === $current_theme_parent->stylesheet );
+				// Detect if current theme parent is the one registered as newer SDK
+				$is_newest_sdk_plugin_active = ( strpos( $fs_newest_sdk->plugin_path, '../' . $themes_directory_name . '/' . $current_theme_parent->stylesheet . '/' ) === 0 );
```
swashata, line 258: Please see my suggestion

start.php
```diff
@@ -309,6 +326,9 @@ function_exists( 'wp_is_json_request' ) &&
 	}
 
 	if ( class_exists( 'Freemius' ) ) {
+		if ($is_theme && $is_current_sdk_newest) {
```
swashata, line 329: Looks like you forgot to remove this part?

swashata (changes requested):

start.php
```diff
@@ -225,25 +242,35 @@ function_exists( 'wp_is_json_request' ) &&
...
+         * True if the plugin with the newest SDK is active.
+         * True if the newest SDK is part of the current theme or current theme's parent.
+         * False otherwise.
+         */
+        if ( ! $is_newest_sdk_type_theme ) {
+            $is_newest_sdk_module_active = is_plugin_active( $fs_newest_sdk->plugin_path );
+        } else {
+            $current_theme = wp_get_theme();
+            // Detect if current theme is the one registered as newer SDK
+            $is_newest_sdk_module_active = ( strpos( $fs_newest_sdk->plugin_path,
```
swashata, line 256:
```php
            $is_newest_sdk_module_active = (
                strpos(
                    $fs_newest_sdk->plugin_path,
                    '/' . $themes_directory_name . '/' . $current_theme->stylesheet . '/'
                ) !== false
            );
```

swashata, line 256: Also do you think a `=== 0` would be a more appropriate check here, like the one for the parent theme? (line 268)

start.php
```diff
@@ -225,25 +242,35 @@ function_exists( 'wp_is_json_request' ) &&
...
 
             $current_theme_parent = $current_theme->parent();
 
             /**
              * If the current theme is a child of the theme that has the newest SDK, this prevents a redirects loop
              * from happening by keeping the SDK info stored in the `fs_active_plugins` option.
              */
-            if ( ! $is_newest_sdk_plugin_active && $current_theme_parent instanceof WP_Theme ) {
-                $is_newest_sdk_plugin_active = ( $fs_newest_sdk->plugin_path === $current_theme_parent->stylesheet );
+            if ( ! $is_newest_sdk_module_active && $current_theme_parent instanceof WP_Theme ) {
```
swashata, line 265:
```php
                // Detect if current theme parent is the one registered as newer SDK
                $is_newest_sdk_module_active = (
                    strpos(
                        $fs_newest_sdk->plugin_path,
                        '../' . $themes_directory_name . '/' . $current_theme_parent->stylesheet . '/'
                    ) === 0
                );
```

start.php
```diff
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.9.0.4';
+	$this_sdk_version = '2.9.0.5';
```
swashata, line 18: @DanieleAlessandra You will find in line 40 a note about theme. Let's add something for your changes too

CleanShot 2024-11-28 at 17 51 01@2x

Something like this:

```
	 * We updated the logic to support SDK loading from a subfolder of a theme as well as from a parent theme.
	 * This allows the SDK to be loaded from composer dependencies or from a custom `vendor/freemius` folder.
	 *
	 * @since 2.9.0.5
	 * @author Daniele Alessandra (@DanielleAlessandra)
```

swashata (commented):
Please take a look at my minor changes here, thanks.

swashata (approved):
Nicely done

PR #759, opened 2024-11-11
Fix PHP 8.2 deprecation warning

DanieleAlessandra:
The report on PHP 8.2 deprecation warnings for dynamic properties highlighted three cases, which this PR addresses:

1. **FS_Plugin_Plan::pricing** – This was resolved in June.
2. **FS_Site::blog_id** – This issue was due to direct assignment.
3. **FS_User::is_beta** – This resulted from the use of `unserialize`.

### Changes:
- Removed the `@property FS_Pricing[] $pricing` annotation from `FS_Plugin_Plan`, as the property now exists and the annotation is no longer necessary.
- Added the `$blog_id` property to `FS_Site` and removed the corresponding annotation, similar to the previous class.
- Introduced the `__wakeup` magic method to `FS_User` to remove deprecated properties from serialized entities.
- Added the `__unserialize` magic method to `FS_User` to manage the result of `unserialize` and prevent warnings. This method only works with public properties and can be removed if the `__wakeup` method is sufficient.


includes/entities/class-fs-user.php
```diff
@@ -48,6 +48,19 @@ function __construct( $user = false ) {
...
 		}
 
+		/**
+		 * @return void
+		 *
+		 * This function removes the deprecated 'is_beta' property from the serialized data.
+		 * Should clean up the serialized data to avoid PHP 8.2 warning on next execution.
+		 */
+		public function __wakeup() {
+			if ( isset($this->is_beta) ) {
```
fajardoleo, line 58: @DanieleAlessandra What if `is_beta` is null?

includes/entities/class-fs-user.php
```diff
@@ -48,6 +48,19 @@ function __construct( $user = false ) {
 			parent::__construct( $user );
 		}
 
+		/**
+		 * @return void
```
fajardoleo, line 52: @DanieleAlessandra I think this should be after the description.

includes/entities/class-fs-user.php
```diff
@@ -48,6 +48,19 @@ function __construct( $user = false ) {
...
+		 *
+		 * This function removes the deprecated 'is_beta' property from the serialized data.
+		 * Should clean up the serialized data to avoid PHP 8.2 warning on next execution.
+		 */
+		public function __wakeup() {
```
fajardoleo, line 57: @DanieleAlessandra When the method is public, there's no need to include `public`.

includes/entities/class-fs-user.php
```diff
@@ -48,6 +48,19 @@ function __construct( $user = false ) {
 			parent::__construct( $user );
 		}
 
+		/**
+		 * @return void
+		 *
+		 * This function removes the deprecated 'is_beta' property from the serialized data.
```
fajardoleo, line 54: @DanieleAlessandra Please use `method` instead of `function`.

includes/entities/class-fs-site.php
```diff
@@ -10,15 +10,15 @@
...
-     * @property int $blog_id
-     */
-    #[AllowDynamicProperties]
     class FS_Site extends FS_Scope_Entity {
         /**
          * @var number
          */
         public $site_id;
+        /**
+         * @var number
```
fajardoleo, line 19: @DanieleAlessandra This should be `int` like in the `get_unfiltered_site_url()` method. For entity IDs though, we use `number` (like in the `FS_Entity` class).

DanieleAlessandra:
This PR needs to be removed, cannot merge directly to master

PR #760, opened 2024-11-12
[parallel-activation] Improved the support for plugins with free and …

fajardoleo:
…premium versions that can be activated in parallel.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -110,6 +110,12 @@
         private $_enable_anonymous = true;
 
         /**
+         * @since 2.9.1
+         * @var string|null Hints the SDK whether the plugin supports parallel activation mode, preventing the auto-deactivation of the free version when the premium version is activated, and vice versa.
+         */
+        private $_premium_plugin_basename_from_parallel_activation;
+
+        /**
          * @since 1.1.7.5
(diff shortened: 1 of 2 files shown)


includes/class-freemius.php
```diff
@@ -5155,19 +5175,43 @@ private function parse_settings( &$plugin_info ) {
...
+            $is_premium     = $this->get_bool_option( $plugin_info, 'is_premium', true );
             $premium_suffix = $this->get_option( $plugin_info, 'premium_suffix', '(Premium)' );
 
+            $module_type = $this->get_option( $plugin_info, 'type', $this->_module_type );
+
+            $parallel_activation = $this->get_option( $plugin_info, 'parallel_activation' );
+
+            if (
+                ! $is_premium &&
+                ( WP_FS__MODULE_TYPE_PLUGIN === $module_type ) &&
```
swashata, line 5187: Micro performance, given we can assume the `plugin` type is most common, yet parallel activation would be fairly uncommon:

```php
if (
  ! $is_premium &&
  is_array( $parallel_activation ) &&
  ...
)
```

swashata (changes requested):

swashata (approved):
Again, nicely done 👏

PR #762, opened 2024-11-22
[cron] Improved the scheduling of the install sync cron job.

fajardoleo:


includes/class-freemius.php
```diff
@@ -6620,10 +6643,6 @@ private function get_install_sync_cron_blog_id() {
          * @param int $except_blog_id Since 2.0.0 when running in a multisite network environment, the cron execution is consolidated. This param allows excluding excluded specified blog ID from being the cron executor.
          */
         private function schedule_install_sync( $except_blog_id = 0 ) {
```
swashata, line 6645: I see that this function is only called from within the `maybe_schedule_install_sync_cron`. So I am suggesting to completely remove this function and move the one line remaining code inside the new function only.

This way we will protect ourselves from accidentally calling this one, which could be harmful as we've seen.

includes/class-freemius.php
```diff
@@ -6514,6 +6514,29 @@ private function maybe_schedule_sync_cron() {
             $this->schedule_sync_cron();
         }
 
+        /**
+         * @author Leo Fajardo (@leorw)
+         * @since  2.9.1
+         */
+        private function maybe_schedule_install_sync_cron() {
```
swashata, line 6521: Let's also propagate all the arguments from the original method `$except_blog_id`

includes/class-freemius.php
```diff
@@ -15981,7 +16000,7 @@ private function update_multisite_data_after_site_deactivation( $context_blog_id
             if ( $this->is_install_sync_scheduled() &&
                  $context_blog_id == $this->get_install_sync_cron_blog_id()
             ) {
-                $this->schedule_install_sync( $context_blog_id );
+                $this->maybe_schedule_install_sync_cron( $context_blog_id );
```
swashata, line 16003: I see the parameter is used indeed.

swashata (changes requested):

swashata (approved):
Nice 💪

PR #764, opened 2024-12-03
Version 2.10.0

swashata:
# Composer Support, Parallel Free/Premium Activation, Pricing Page Update, and Various Fixes

We are glad to announce **v2.10.0** of our WordPress SDK. This version introduces several improvements and bug fixes. Please find the details below.

## Composer Support and WP.org Compliance

The WordPress.org review team now requires our SDK to be loaded from a directory named `vendor` to ensure compliance with repository standards.

Until now, we asked our makers to download and place the SDK inside the root directory of their product and rename the directory to `freemius`. 

To comply with this requirement, we’ve updated our SDK to be loaded from the `vendor/freemius` directory instead. You can manually copy the SDK into such a directory structure, and it will work seamlessly for both plugins and themes.

vendor-directory-compliance

Please note that you will also need to update the integration snippet to load the SDK `start.php` from the new directory and we recommend "committing" the change so that the reviewer team can see it.

```php
require_once dirname( __FILE__ ) . '/vendor/freemius/start.php';
```

We also recognize the growing need for Composer-based installations, especially since many modern plugins and themes use this dependency manager. To enhance developer experience (DX), we now officially support Composer. 

You can install our SDK with the following command:

```sh
composer require freemius/wordpress-sdk
```

Then, require the Composer autoloader file:

```php
require_once __DIR__ . '/vendor/autoload.php';
```

Our SDK will work out of the box without any further modifications.

Since this is a compliance request from WordPress.org, we strongly encourage you to make the necessary changes as soon as possible.

## Parallel Activation of Both Free and Premium Versions

Historically, our SDK supported activating either the Free or the Premium version of a WordPress plugin. It also included a mechanism to automatically deactivate the free version when the premium version was activated.

However, some use cases require both free and premium versions to remain active, particularly to collect better statistics from WordPress.org.

Previously, we introduced the `deactivate_on_activation` filter to modify this behavior. However, it wasn’t a complete solution and could lead to bugs in scenarios like bulk plugin activation or updates on WordPress Multisite networks.

We’ve now completed this feature by introducing a new initialization code. To use this feature, add the following code to your `fs_dynamic_init` call:

```php
$my_fs = fs_dynamic_init( [
    // ... existing configuration options.
    'parallel_activation' => [
        'enabled'                  => true,
        'premium_version_basename' => 'premium-slug/filename.php',
    ],
] );
```

With this code, the SDK will:

1. Avoid deactivating the free version during premium version activation (unless specifically modified by the `deactivate_on_activation` filter).
2. Automatically show the license activation page if the free version does not have a license activated.

If you encounter any issues using this feature, please contact our support team.

## Pricing Page Update

pricing-page-1 1 0

The SDK now includes [v1.1.0](https://github.com/Freemius/pricing-page/releases/tag/1.1.0) of the pricing page app. Based on community feedback, we’ve introduced the following enhancements:

- [x] Added the `pricing/show_annual_in_monthly` filter to control how annual billing cycle pricing is displayed.
- [x] Introduced the `plugin_icon` filter to override the image displayed on the page.
- [x] Added the `pricing/disable_single_package` filter to control the rendering of single-plan products.
- [x] Introduced the `pricing/css_path` filter to add custom CSS to the pricing page.

More details can be found in the [README](https://github.com/Freemius/pricing-page?tab=readme-ov-file#filters) file of the pricing page repository.

Additionally, we’ve fixed the following issues:

1. The "Lifetime" tab was incorrectly displayed in some cases. This has been resolved.
2. The "Free" plan was mistakenly rendered with an "Upgrade" button, even when it was the current plan. This has been fixed.
3. The current plan was not properly highlighted in some scenarios. This has been corrected.

## Other bug fixes and enhancements

* Improved heuristics for cron-based syncing of license and install entities to avoid performance issues, especially on Multisite networks.
* The SDK now automatically detects many sandbox/playground sites (e.g., [**WordPress Playground**](https://playground.wordpress.net/), InstaWP, TasteWP) and enables anonymous mode to skip the opt-in process.
* Fixed an edge-case JavaScript issue that caused the license activation dialog to fail in some scenarios.
* Updated staging site URL pattern for Kinsta.


vovafeldman:
Amazing release 👏

PR #767, opened 2024-12-05
Use Theme name as 'plugin_path'

DanieleAlessandra:


swashata:
Already managed through https://github.com/Freemius/wordpress-sdk/pull/769

PR #768, opened 2024-12-05
[playground] [anonymous-mode] Fixed the WP Playground anonymous mode …

fajardoleo:
…support.


swashata:
Closing because it will be included in https://github.com/Freemius/wordpress-sdk/pull/769

PR #773, opened 2024-12-12
Add type to composer.json

sanzeeb3:
It would allow the `composer/installers` to install the package to a directory other than the default `vendor` directory.


swashata:
Hello @sanzeeb3 current WordPress.org guideline now requires our SDK to be installed inside a directory named `vendor`. Also from my understanding the default [type](https://getcomposer.org/doc/04-schema.md#type) is a `library`. 

CleanShot 2025-04-30 at 11 01 06@2x


So could you please tell me what other benefit the explicit `type` has? Is it not redundant?

sanzeeb3:
@swashata - looks like that. It needs to be other than the `library` for the custom installer paths to work.

PR #779, opened 2024-12-29
[menu] Added a new filter for enabling the advanced CPT-related menu …

fajardoleo:
…logic.

### DIFF
--- a/includes/managers/class-fs-admin-menu-manager.php
+++ b/includes/managers/class-fs-admin-menu-manager.php
@@ -700,21 +700,27 @@
 			}
 
 			$menu_slug   = $menu['menu'][2];
-			$parent_slug = 'admin.php';
+			$parent_slug = isset( $menu['parent_slug'] ) ?
+				$menu['parent_slug'] :
+				'admin.php';
 
-			/**
-			 * This line and the `if` block below it is based on the `menu_page_url()` function of WordPress.
-			 *
-			 * @author Leo Fajardo (@leorw)
-			 * @since 2.7.4
-			 */
-			global $_parent_pages;
+			if ( fs_apply_filter( $this->_module_unique_affix, 'enable_cpt_advanced_menu_logic', false ) ) {
+				$parent_slug = 'admin.php';
 
-			if ( ! empty( $_parent_pages[ $menu_slug ] ) ) {
-				$_parent_slug = $_parent_pages[ $menu_slug ];
-				$parent_slug  = isset( $_parent_pages[ $_parent_slug ] ) ?
-					$parent_slug :
-					$menu['parent_slug'];
+				/**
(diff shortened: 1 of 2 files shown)


includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -700,21 +700,27 @@ function main_menu_url() {
...
-				$_parent_slug = $_parent_pages[ $menu_slug ];
-				$parent_slug  = isset( $_parent_pages[ $_parent_slug ] ) ?
-					$parent_slug :
-					$menu['parent_slug'];
+			$parent_slug = isset( $menu['parent_slug'] ) ?
+				$menu['parent_slug'] :
+				'admin.php';
+
+			if ( fs_apply_filter( $this->_module_unique_affix, 'enable_cpt_advanced_menu_logic', false ) ) {
+				$parent_slug = 'admin.php';
```
DanieleAlessandra, line 708: Why do we need to re-define `$parent_slug` here? it has just been defined on line 703.

includes/managers/class-fs-admin-menu-manager.php
```diff
@@ -700,21 +700,27 @@ function main_menu_url() {
...
-			if ( ! empty( $_parent_pages[ $menu_slug ] ) ) {
-				$_parent_slug = $_parent_pages[ $menu_slug ];
-				$parent_slug  = isset( $_parent_pages[ $_parent_slug ] ) ?
-					$parent_slug :
-					$menu['parent_slug'];
+			$parent_slug = isset( $menu['parent_slug'] ) ?
+				$menu['parent_slug'] :
+				'admin.php';
+
+			if ( fs_apply_filter( $this->_module_unique_affix, 'enable_cpt_advanced_menu_logic', false ) ) {
```
DanieleAlessandra, line 707: Do we have documentation about `enable_cpt_advanced_menu_logic` hook somewhere?

DanieleAlessandra (changes requested):
Please take a look to my comments

includes/managers/class-fs-admin-menu-manager.php
fajardoleo: @DanieleAlessandra The line above preserves the original behavior when the new filter is not used. When the filter is applied, the behavior changes slightly, and not using `admin.php` may lead to unexpected issues.

includes/managers/class-fs-admin-menu-manager.php
fajardoleo: @DanieleAlessandra This will be documented. Thanks!

DanieleAlessandra (approved):
Looks good to me.

swashata (approved):
LGTM 👏

PR #789, opened 2025-02-28
Fix: Prevent SDK from exiting in CLI environments

DanieleAlessandra:
Changed `exit;` to `return;` in `start.php` to prevent the Freemius SDK from terminating execution in CLI environments where `ABSPATH` is not defined. This resolves conflicts with tools like PHPCS, PHPStan, and PHPUnit while ensuring the SDK remains functional within WordPress.

### DIFF
--- a/start.php
+++ b/start.php
@@ -7,7 +7,7 @@
 	 */
 
 	if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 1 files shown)


fajardoleo (approved):
Looks good to me.

PR #791, opened 2025-03-06
Always hook the data sync cron event callback

johnbillion:
Currently the cron event callback to sync data is only hooked when the cron runner is actually running. This causes a few problems:

* A cron management plugin such as WP Crontrol will flag the event as having no action.
* Using an alternative cron runner such as Cavalcade or Cron Control will result in the action firing but the callback not being hooked. The callback will never get called.

There is no need to conditionally hook the cron event callback only when cron is running. This fixes that.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -4495,9 +4495,9 @@
             }
 
             if ( $this->has_api_connectivity() ) {
-                if ( self::is_cron() ) {
-                    $this->hook_callback_to_sync_cron();
-                } else if ( $this->is_user_in_admin() ) {
+                $this->hook_callback_to_sync_cron();
+
+                if ( $this->is_user_in_admin() ) {
                     /**
                      * Schedule daily data sync cron if:
                      *


includes/class-freemius.php
```diff
@@ -4495,9 +4495,9 @@ function dynamic_init( array $plugin_info ) {
             }
 
             if ( $this->has_api_connectivity() ) {
-                if ( self::is_cron() ) {
-                    $this->hook_callback_to_sync_cron();
-                } else if ( $this->is_user_in_admin() ) {
+                $this->hook_callback_to_sync_cron();
+
+                if ( $this->is_user_in_admin() ) {
```
swashata, line 4500: Let's do this just to be safe about the logic change

```suggestion
                if ( ! self::is_cron() && $this->is_user_in_admin() ) {

```

swashata:
Thanks @johnbillion the changes needed a bit more polishing. We would be doing it here https://github.com/Freemius/wordpress-sdk/pull/808

PR #792, opened 2025-03-12
[privacy] Avoid sending first name, last name and email if not needed

DanieleAlessandra:
To improve data efficiency and user experience, this change adjusts the activation form. It prevents the submission of `firstname`, `lastname`, and `email` when a license is already linked to a user. When a user is required for license activation, these fields are sent, and a clear notification is provided to the user.


fajardoleo (approved):
Looks good to me.

PR #797, opened 2025-04-29
[starter] [themes] Improved the "calculation" of the themes directory…

fajardoleo:
… by normalizing the calculated path since some environments can return inconsistent slash usages.


swashata (approved):
Nice. I see that we are also normalizing the `$file_path` too.

PR #802, opened 2025-05-18
[multisite] Improve retrieval of individual‑site info in a multisite …

fajardoleo:
…network.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -7814,9 +7814,9 @@
          *
          * @param FS_Plugin_License $license
          * @param array             $sites
-         * @param int               $current_blog_id
+         * @param int               $blog_id
          */
-        private function activate_bundle_license( $license, $sites = array(), $current_blog_id = 0 ) {
+        private function activate_bundle_license( $license, $sites = array(), $blog_id = 0 ) {
             $is_network_admin = fs_is_network_admin();
 
             $installs_by_blog_map  = array();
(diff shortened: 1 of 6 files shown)


includes/class-freemius.php
```diff
@@ -7897,18 +7897,20 @@ private function activate_bundle_license( $license, $sites = array(), $current_b
                         }
                     }
 
+                    $current_blog_id = get_current_blog_id();
```
swashata: I am not fully aware of the mechanism, but don't we want to call this before even switching on line `7854`?

includes/class-fs-hook-snapshot.php
```diff
@@ -0,0 +1,45 @@
...
+
+		private $removed_actions = array();
+
+		/**
+		 * Remove all actions from a given hook and store them for later restoration.
+		 */
+		public function remove( $hook ) {
+			global $wp_filter;
+
+			if ( isset( $wp_filter[ $hook ] ) ) {
```
DanieleAlessandra: In normal WP context this is good enough, but in edge cases it could be useful to safe check the existence of `$wp_filter`: 

```if ( isset( $wp_filter ) &&  isset( $wp_filter[ $hook ] ) ) {```

 This makes the code more robust in edge cases, such as when the file is loaded outside of the standard WordPress environment or during automated tests.
In normal WordPress usage, this is rarely necessary, but it helps prevent potential errors in atypical scenarios.

**Review by DanieleAlessandra — APPROVED** [review]
The changes look good enough to me, only read the one comment and think about it (we have to deal with tons of different use cases).

File: includes/class-freemius.php [hunk 3]
```diff
@@ -15832,9 +15833,8 @@ function get_site_info( $site = null, $load_registration = false ) {
                 $info[ 'registration_date' ] = $registration_date;
             }
 
-            if ( $switched ) {
```
swashata: @fajardoleo why are we removing this? Don't we want to restore the blog like before?

swashata (commented):
Please see my only question

swashata (approved):
Nice addition

PR #814, opened 2025-07-10
[deactivation-feedback] Improve subscription cancellation panel copy …

fajardoleo:
…in the plugin deactivation feedback dialog.


DanieleAlessandra (approved):
Got it

PR #816, opened 2025-07-16
[debug] [api] Improve the wording of the API status on the debug page…

fajardoleo:
… to prevent confusion.

### DIFF
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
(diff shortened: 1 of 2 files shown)


DanieleAlessandra (approved):
It works and makes sense — the only thing I would have changed is not using the green background for the row, but I didn’t find any specific instructions about it in the requirements.

DanieleAlessandra (approved):
Better

PR #817, opened 2025-08-11
[php82] [php-error]  Remove the skype_support dynamic property creation php error. Adds missing property from serialized object for the support_skype requirement.

bahiirwa:
- [x] Remove the dynamic property creation php error. 
- [x] Adds missing property from serialized object for the support_skype requirement.


fajardoleo (approved):
Looks good to me.

PR #821, opened 2025-08-31
[multisite] [cache] Fix the site-activated plugins cache so that it o…

fajardoleo:
…nly has the relevant properties.


DanieleAlessandra (approved):
Looks good: we only keep `slug`, `Version` and `name`, like we do on lines 8885-8887

swashata (approved):
Nice

PR #825, opened 2025-09-23
Code Modernization: Address __wakeup() deprecation in PHP 8.5.

TobiasBg:
PHP 8.5 deprecates the `__sleep()` and `__wakeup()` magic methods in favor of `__serialize()` and `__unserialize()`.

See https://wiki.php.net/rfc/deprecations_php_8_5#deprecate_the_sleep_and_wakeup_magic_methods

For PHP < 7.4 compatibility, `__sleep()` and `__wakeup()` need to be kept for the time being.

This commit moves the logic of the `__wakeup()` method to `__unserialize()`, and turns the former into a wrapper. The SDK does not use `__sleep()` methods, so this is the only change required.

### DIFF
--- a/includes/entities/class-fs-user.php
+++ b/includes/entities/class-fs-user.php
@@ -54,11 +54,18 @@
 		 *
 		 * @return void
 		 */
-		function __wakeup() {
+		function __unserialize( $data ) {
 			if ( property_exists( $this, 'is_beta' ) ) {
 				// If we enter here, and we are running PHP 8.2, we already had the warning. But we sanitize data for next execution.
 				unset( $this->is_beta );
 			}
+		}
+
+		/**
+		 * Compatibility wrapper of `__unserialize` for PHP < 7.4.
+		 */
+		function __wakeup() {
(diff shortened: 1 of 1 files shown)


TobiasBg:
Ok, this might need some more code to actually restore properties, see the ongoing discussion at https://core.trac.wordpress.org/ticket/63962#comment:11 and following comments.

swashata:
@TobiasBg looks like the deprecation was cancelled (like you said so in the ticket) https://wiki.php.net/rfc/soft-deprecate-sleep-wakeup

And we tried out an implementation here in https://github.com/Freemius/wordpress-sdk/pull/831 The implementation worked but it produced more boilerplate code (the exact reason for which the deprecation was cancelled).

So I am also closing this as `wontfix` for now. If the deprecation ever becomes real we can revisit.

TobiasBg:
Hi @swashata!
Yes, good call, this can be `wontfix`ed with the deprecation having been removed.
Totally forgot to update/close this PR as well, sorry.

PR #826, opened 2025-09-29
[action] [license-deactivation] Add a new action that is triggered af…

fajardoleo:
…ter a license deactivation.


DanieleAlessandra (approved):
Tested all flows (manual deactivation, downgrade with deactivate_license, expired license). Hook always fired as expected. ✅

PR #828, opened 2025-10-14
[action] [license-activation] Add a new action that is triggered afte…

fajardoleo:
…r a license activation.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -14034,6 +14034,10 @@
                 $result['next_page'] = $next_page;
             }
 
+            if ( $result['success'] ) {
+                $this->do_action( 'after_license_activation' );
+            }
+
             return $result;
         }
 
@@ -21666,6 +21670,8 @@
 
                 return;
             }
+
+            $this->do_action( 'after_license_activation' );
 
             $premium_license = new FS_Plugin_License( $license );
 
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.12.2.1';
+	$this_sdk_version = '2.12.2.2';
 
 	#region SDK Selection Logic --------------------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


includes/class-freemius.php
```diff
@@ -14034,6 +14034,8 @@ private function activate_license(
                 $result['next_page'] = $next_page;
             }
 
+            $this->do_action( 'after_license_activation', ( true === $result['success'] ) );
```
swashata: How about we keep it simple now and just trigger the action if the activation was successful? Like

```php
if ( $result['success'] ) {
    $this->do_action...
}
```

will that work?

includes/class-freemius.php
```diff
@@ -21570,6 +21572,8 @@ public function _open_license_activation_dialog_box() {
          *
          * @param bool                   $background
          * @param FS_Plugin_License|null $premium_license
+         *
+         * @return bool
          */
         protected function _activate_license( $background = false, $premium_license = null ) {
```
swashata, line 21578: @fajardoleo how about you revert all these changes (and also line `23080` inside `_handle_account_edits()`) and at line `21673`, before the `$premium_license = new FS_Plugin_License( $license );` you fire the action directly?

```
$this->do_action( 'after_license_activation' );
```

again we fire it only when the license is actually activated. No need to pass a param suggesting the "status" of it, just to keep things simple and straightforward. (Conceptually one may not trigger an action that happens after license activation if the license wasn't activated to begin with ;) )

includes/class-freemius.php
```diff
@@ -23071,7 +23077,9 @@ private function _handle_account_edits() {
                     }
 
                     if ( is_object( $fs ) ) {
-                        $fs->_activate_license();
+                        $result = $fs->_activate_license();
```
swashata: Please see my suggestion, perhaps we won't be needing this change too.

swashata (changes requested):

swashata (approved):

PR #829, opened 2025-10-16
[payments] Add VAT to the gross amount for matching values in the accounts page in SDK

bahiirwa:
- [x] Add VAT to the gross amount to ensure the accounts page in SDK has matching values to the customer portal and invoice.

### DIFF
--- a/includes/entities/class-fs-payment.php
+++ b/includes/entities/class-fs-payment.php
@@ -132,10 +132,11 @@
          */
         function formatted_gross()
         {
+            $price = $this->gross + $this->vat;
             return (
-                ( $this->gross < 0 ? '-' : '' ) .
+                ( $price < 0 ? '-' : '' ) .
                 $this->get_symbol() .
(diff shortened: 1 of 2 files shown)


includes/entities/class-fs-payment.php
```diff
@@ -132,10 +132,11 @@ function is_migrated() {
          */
         function formatted_gross()
         {
+            $price = round($this->gross + $this->vat, 2);
```
fajardoleo: @bahiirwa Missing spaces, but do we still need `round()` here? (We're already rounding the value when calling `number_format()`.)

fajardoleo (approved):
Looks great.

includes/entities/class-fs-payment.php
bahiirwa: Considering we do not have a great need for precision like a DB save and only one comparison operation, we can get rid of `round()`.

PR #844, opened 2026-01-11
[debug] [cron] Enrich the debug page with the default WP-Cron status.

fajardoleo:


DanieleAlessandra (approved):
LGTM 👍

The logic is correct and aligned with WordPress standards (DISABLE_WP_CRON is the right signal to check).

Noticed that the value is stored as 'true'/'false' strings instead of booleans, but since the whole debug table already follows this pattern, it’s consistent with the existing implementation.

Approved as-is.

PR #845, opened 2026-01-13
Remove json2

### LINKED ISSUE #840: Remove `json2` script dependency.
WP 6.9 deprecated the `json2` script dependency -- it's only relevant for very old versions of IE that are no longer supported by anyone.

The SDK enqueues the `json2` dependency, which now triggers a deprecation notice.
To get rid of that the `json2` script should no longer be enqueued.

johnbillion:
In WordPress 6.9, the `json2` library was no-oped as it's no longer needed by any browser. This was done as part of a larger change to remove support for old IE conditional syntax. See https://core.trac.wordpress.org/ticket/63821 and https://github.com/WordPress/wordpress-develop/commit/49d1dedd7b7efbe6f783931ebbf84e7931afb937.

This PR removes `json2` from the Freemius SDK. This should be a safe change to make because:

- If a Freemius-powered plugin depends on `json2` then it'll still get enqueued via the normal script dependency system in WordPress. The `json2` script is still registered, it's just no longer enqueued manually by the Freemius SDK.
- The `json2` library provides polyfills for `JSON.stringify()` and `JSON.parse()` by pushing its implementations onto the global `JSON` object. No browsers need this back compat, and there is nothing provided in `json2` which is not provided by `JSON` in the browser.

Fixes #840.

### DIFF
--- a/includes/class-freemius.php
+++ b/includes/class-freemius.php
@@ -7098,7 +7098,6 @@
          */
         function _enqueue_connect_essentials() {
             wp_enqueue_script( 'jquery' );
-            wp_enqueue_script( 'json2' );
 
             fs_enqueue_local_script( 'postmessage', 'nojquery.ba-postmessage.js' );
             fs_enqueue_local_script( 'fs-postmessage', 'postmessage.js' );
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
-	$this_sdk_version = '2.13.0.2';
+	$this_sdk_version = '2.13.0.3';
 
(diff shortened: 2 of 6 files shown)


bahiirwa (approved):

DanieleAlessandra (approved):
Thanks for the clear explanation and for referencing the WordPress core change.
The reasoning makes sense and the implementation looks solid to me.
Removing the manual enqueue of json2 in the SDK is safe and aligns with WordPress 6.9 behavior.

LGTM 👍

bahiirwa:
Thanks @johnbillion and @DanieleAlessandra 
I made a force push to fix the commit history so as to match the develop branch.

PR #847, opened 2026-01-20
[deprecation] [warning] Remove using null as an array offset is deprecated and will trigger a deprecation notice in PHP 8.5.

bahiirwa:
**REASON:** 
In PHP 8.5, using null as an array offset is deprecated and will trigger a deprecation notice. Previously, null was silently coerced to an empty string ("") when used as an array key, but this implicit behavior is being phased out to improve consistency with other parts of the language. 

**Solution**
Changing the null to empty strings fixed the issue, as the WordPress function expects strings to be used as URI/URLs. https://developer.wordpress.org/reference/functions/add_query_arg/

image

### DIFF
--- a/includes/managers/class-fs-contact-form-manager.php
+++ b/includes/managers/class-fs-contact-form-manager.php
@@ -77,7 +77,7 @@
 			$query_params = $this->get_query_params( $fs );
 
 			$query_params['is_standalone'] = 'true';
-			$query_params['parent_url']    = admin_url( add_query_arg( null, null ) );
+			$query_params['parent_url']    = admin_url( add_query_arg( '', '' ) );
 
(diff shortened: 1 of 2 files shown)


DanieleFreemius (approved):
LGTM

Passing empty strings to `add_query_arg()` is acceptable here.
While using `add_query_arg( [], admin_url() )` would be semantically cleaner, the current change preserves existing behavior with a minimal diff and avoids the PHP 8.5 deprecation without introducing any functional or stylistic refactors.

Approved.

DanieleAlessandra (approved):
LGTM

Passing empty strings to `add_query_arg()` is acceptable here.
While using `add_query_arg( [], admin_url() )` would be semantically cleaner, the current change preserves existing behavior with a minimal diff and avoids the PHP 8.5 deprecation without introducing any functional or stylistic refactors.

Approved.

PR #851, opened 2026-02-27
[sync-cron] Add a grace period before rescheduling the data sync cron…

fajardoleo:
… event to prevent a race condition between the rescheduling process and the execution of the scheduled event.


swashata (approved):
LGTM

PR #867, opened 2026-06-19
[api] [timeout] …

DanieleAlessandra:
Introduced a constant for API timeout and updated usage in the SDK

### DIFF
--- a/config.php
+++ b/config.php
@@ -210,8 +210,8 @@
         define( 'WP_FS__API_SANDBOX_ADDRESS_LOCALHOST', 'http://sandbox-api.freemius:8080' );
     }
 
-    if ( ! defined( 'WP_FS_API__TIMEOUT' ) ) {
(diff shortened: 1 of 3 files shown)


config.php
```diff
@@ -210,6 +210,10 @@
         define( 'WP_FS__API_SANDBOX_ADDRESS_LOCALHOST', 'http://sandbox-api.freemius:8080' );
     }
 
+    if ( ! defined( 'WP_FS_API__TIMEOUT' ) ) {
+        define( 'WP_FS_API__TIMEOUT', 30 );
```
fajardoleo: @DanieleAlessandra For consistency, I suggest `WP_FS__API_TIMEOUT` (like the other constants).

fajardoleo (changes requested):

config.php
DanieleAlessandra: Done

fajardoleo (approved):
Looks great.

PR #883, opened 2026-08-05
[addons]  Ensure cheapest addons pricing is displayed across all plans and cycles.

bahiirwa:
- [x] Ensure cheapest pricing is displayed across all plans and cycles.
- [x] Including lifetime pricing in the calculation when only the lifetime pricing has been set for the product.

### DIFF
--- a/start.php
+++ b/start.php
@@ -15,7 +15,7 @@
 	 *
 	 * @var string
 	 */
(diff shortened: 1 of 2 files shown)


vienlhanz26 (approved):
Looks good.

